PR #17, opened 2019-10-16
OAuth module

awichera:
So, there are two modules added: authentication, which I would probably remove, But It makes sense to me to back it up somewhere, it may come handy. It may be eventualty devoured by Seacat Auth or something.

Oauth module is what this is about.

The code is a mess at the moment and is in need of a refactorization. I would appreciate suggestions for it, hence this merge request.

__________________________________________________________
Ideas:

- Move all the logic from from components to AuthService
- Create authProvider json, which would be used to draw buttons on /auth screen. It could work well with some config, that would then filter only wanted authProviders.
- I like specific flow files for specific oAuth providers, adding new one was quite simple
- As dicussed on last review, change indentity handling to 
```load app => check for token in cookie => validate token on the server, get user info => continue as logged in/out user```

Questions:
- How to catch errors in requests, and don't make the code too messy?
- Where exactly is it customary to separate view and controller?
- Views shold be obviously container components, and controllers? Services? Single service? Something else?
- What is supposed to be in module configuration?
     - some form of list of oAuth providers
     - what else?

### CODE REVIEW

File: src/modules/oauth/index.js [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+import TeskalabsAuth from './containers/TeskalabsAuth'
+import GitHubAuth from './containers/GitHubAuth'
+import AuthService from './services/AuthService'
+
+
+
+export default class AuthenticationModule extends Module {
+    constructor(app, name){
+        super(app, "Authentication");
+
```
ateska: Meh ... don't hardcode that vendor-specific things (TL Auth, GitHub) ... it shouldn't be part of the library.
On the other hand, it should be captured in examples/docs/

src/containers/Application.js
```diff
@@ -38,15 +38,20 @@ class Application extends Component {
 
 
 		// Instantiate modules
+		console.log ("IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII")
```
PremyslCerny: Be more specific about the log message, which component it is located in etc. Obviously, user will not want to see it unless there is some error.

src/modules/authentication/containers/RegisterBox.js
```diff
@@ -0,0 +1,274 @@
...
+      this.setState({ email });
+      this.doValidateSubmit ()
+    }
+
+    handlePwdInput(event) {
+      const { pwd, repPwd} = this.state
+      pwd.value = event.target.value
+      this.setState({ pwd });
+      if (pwd.validate) {
+        this.doValidatePwd (event.target.value);
```
PremyslCerny: Remove extra space before the parenthesis.

src/modules/oauth/containers/AuthContainer copy.js
```diff
@@ -0,0 +1,113 @@
+import React, { Component } from 'react'
```
ateska: `AuthContainer copy.js` ?

src/modules/oauth/containers/AuthContainer.js
```diff
@@ -0,0 +1,130 @@
...
+    constructor(props) {
+      super(props);
+      this.AuthService =  this.props.app.locateService("AuthService");
+
+
+      this.state = {
+        buttons:[]
+      };
+
+      const TLdomain = "https://via.teskalabs.com/seacat-auth";
```
ateska: This is too specific to be part of the  generic technology.

src/modules/oauth/containers/ServerInteractionBox.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react'
+import { connect } from 'react-redux';
+import { withRouter } from "react-router";
+import {
+  Alert, Button, Card, CardBody, Col, ListGroup, ListGroupItem, Label,
+  Form, FormGroup, Input, InputGroup, InputGroupAddon, InputGroupText, Row
+} from 'reactstrap';
+import axios from 'axios'
+
+class ServerInteractionBox extends Component {
```
ateska: What is "server interaction"?

awichera:
@ateska It is still WIP. Most of the think you mention in requested changes are to be factored out (ServerInteractionBox, AuthContainer copy.js, links in AuthContainer).

I requested the review because of the question about the architecture - regarding specification of availible authMethods in user code.

src/modules/oauth/containers/AuthContainer copy.js
awichera: Will factor out.

src/modules/oauth/containers/AuthContainer.js
awichera: This will be factored out, url will be in specific authMethods.

src/modules/oauth/containers/ServerInteractionBox.js
awichera: This won't make a sense after refactorization is finished. I will remove it.

ateska:
@awichera - pls, resolve conflicts.

src/modules/oauth/services/AuthService.js
```diff
@@ -0,0 +1,59 @@
+import { connect } from 'react-redux';
+import axios from 'axios';
+import queryString from 'query-string'
+import Service from '../../../abc/Service';
+
+
+
+export default class AuthService extends Service {
```
ateska: There is already `class AuthService` defined ... this is rather `OAuthService`?

src/modules/oauth/containers/GitHubAuth.js
```diff
@@ -0,0 +1,118 @@
...
+import axios from 'axios'
+
+class GitHubAuth extends React.Component {
+
+    constructor(props) {
+      super(props);
+      this.AuthService =  this.props.app.locateService("AuthService");
+    }
+
+  //   async componentDidMount() {
```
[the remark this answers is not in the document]
awichera: Delete it :)

awichera:
I think I am almost at the point of serious merge request. 

I have added a demo, to show my proposal of how the module would be used in a user code in this iterations.
I don't think architecture of ASAB UI allows for extensive (or any) module configuration and I would definiteliy wouldn't drag introduction of this feature in to this merge request.

Given the circumstance, I think that creation of inheriting module with around five lines of code is reasonable way to go about it. At least for the time being.
__________________________

Regarding the comments, I will try to add comments explaining a usage and structure of basic classes.
Any other remarks?

src/modules/oauth/services/AuthService.js
awichera: Maybe?
My idea was, that rest of the application would be apologetic on type of authorization. 
Any form of authorization module would expose AuthService and you could basicly plug and play any type of authorization module in to your app without changing a thing.

Thats why I am using generic name AuthorizationModule as well.

Is that a wrong way to think about it?

ateska: Name has to be unique ... you are mixing type/class and name.
Simply said, there should be only one class of a given name.

ateska: Ok, this seems to be a last bit prior merge.

ateska:
Don't forget to resolve that conflict.

awichera:
@ateska : I somehow thought this has been merged already.
Is there anything else for this request, that should be done, or should I merge it?

PR #18, opened 2019-12-01
1912 refactoring of the ASAB Web UI kit

ateska:


demo/src/index.js
```diff
@@ -18,7 +17,5 @@ modules.push(SampleModule);
 
 // Render
 ReactDOM.render((
-	<BrowserRouter>
-		<Application modules={modules} />
-	</BrowserRouter>
+	<Application modules={modules} />
```
mpavelka: This is ok as long as you don't need route information in the application object. You won't be able to use `withRouter(Application)` AFAIK.

PR #19, opened 2020-02-11
Authmodule refactor

awichera:
Moves user creds from redux store to browser local storage.

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -38,10 +38,6 @@ class Application extends Component {
 
 
 		// Instantiate modules
-		console.log ("IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII")
-		console.log ("INITIATING MODULES")
-		console.log ("IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII")
-
 		for (var i in props.modules) {
 			const module = new props.modules[i](this);
 			this.Modules.push(module);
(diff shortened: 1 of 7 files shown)


PremyslCerny:
Nice improvement

ateska:
I was actually thinking why we do popup ... instead of a full-screen login (redirect & callback URL).
Any feedback here? @awichera @PremyslCerny

awichera:
Well, to be honest,  I don't know, but it feels very customary with oAuth.
Almost everywhere where I encounter oAuth, it is a pop-up.

For me as a user, it kind of feels like it is a separate thing - I am not providing my password to resource owener, but I am communicating directly with auth server.

But, thats very.. behavioral reason, to be sure.

ateska:
Well, I actually met few sites that does page-transition instead of pop-up.
The main reason for this topic is how we will support login via Google/GitHub etc. ... and this is exploratory question - if we can do that in popup seamlessly, I'm fine with that answer.

awichera:
Yeah, we do that in pupup quite seamlessly. in my opinion - well, with GitHub. We don't have GoogleAuthMethod yet.

ateska:
Ok. I would like to see that ... eventually.
Re. my work on SeaCat Auth now.

ateska:
Review is done from my side (FYI).

awichera:
By the way these are some follow up tasks I came up with:
- fix seacat-auth - 500 internal when trying to add user

- improve asab auth module - `/identity` should not reply with 200 if it got 401 from auth provider (Github)
- update TeskalabsAuthMethod - use Github auth as template
- update LDAP Auth Method - use Github auth as template
- ? create wrapper around authorized axios request, that would handle tokens, unauthorized errors and such, to not repeat this functionality with every specific request?

ateska:
I work on a very significant refactoring of SeaCat Auth ... many of above point will become irrelevant. I suggest to reiterate once merged.

awichera:
Sounds like a plan.

PR #26, opened 2020-04-18
Update to a recent ReactStrap / CoreUI changes.

ateska:


src/containers/Application.js
```diff
@@ -122,61 +124,37 @@ class Application extends Component {
...
-									)}
-								/>
-							) : (null);
-						})}
-					</Switch>
+												)}
+											/>
+										) : (null);
+									})}
+									(this.DefaultPath != undefined ? <Redirect from="/" to={this.DefaultPath} />: null)
```
AituAbdiluly, line 147: Instead of parenthesis should be brackets

ateska: Fixed

PR #29, opened 2020-04-23
Add optional sidebar and breadcrumb

Pe5h4:
This pull request allows to disable sidebar and breadcrumb. It also remove items added to header, when the screen is diminished (e.g. in mobile phones). Those header items then appear in the sidebar, if added previously to the Navigation. Example of settings is written in the description of Application.js

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -167,7 +167,9 @@ class Application extends Component {
 				<div className="app">
 					<Header hasSidebar={this.props.hasSidebar} app={this}/>
 					<div className="app-body">
-						<Sidebar hasSidebar={this.props.hasSidebar} app={this} navigation={this.Navigation}/>
+						{(this.props.hasSidebar || typeof this.props.hasSidebar === 'undefined') ? 
+							<Sidebar app={this} navigation={this.Navigation} display="lg"/> : 
+							<Sidebar app={this} navigation={this.Navigation} display="xs"/>}
 						<main className="main">
 							{(this.props.hasBreadcrumb || typeof this.props.hasBreadcrumb === 'undefined') ? 
 								<AppBreadcrumb appRoutes={this.Router.Routes} router={router}/> : null}
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -124,11 +124,12 @@ class Application extends Component {
 		else return (
 			<Provider store={this.Store}>
 				<div className="app">
-					<Header app={this}/>
+					<Header hasSidebar={this.props.hasSidebar} app={this}/>
 					<div className="app-body">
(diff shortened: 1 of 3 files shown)


src/containers/Application.js
```diff
@@ -124,11 +165,12 @@ class Application extends Component {
 		else return (
 			<Provider store={this.Store}>
 				<div className="app">
-					<Header app={this}/>
+					<Header hasSidebar={this.props.hasSidebar} app={this}/>
 					<div className="app-body">
-						<Sidebar app={this} navigation={this.Navigation}/>
+						<Sidebar hasSidebar={this.props.hasSidebar} app={this} navigation={this.Navigation}/>
```
ateska, line 170: The actual use of sidebar can be conditional - not via `hasSidebar` property.

Pe5h4: Changed in last 2 commits

src/containers/Application.js
```diff
@@ -124,11 +165,14 @@ class Application extends Component {
 		else return (
 			<Provider store={this.Store}>
 				<div className="app">
-					<Header app={this}/>
+					<Header hasSidebar={this.props.hasSidebar} app={this}/>
```
ateska, line 168: You can access `app.props.hasSidebar` from Header.

No need to pass `hasSidebar={this.props.hasSidebar}` there.

src/containers/Application.js
```diff
@@ -124,11 +165,14 @@ class Application extends Component {
...
 					<div className="app-body">
-						<Sidebar app={this} navigation={this.Navigation}/>
+						{(this.props.hasSidebar || typeof this.props.hasSidebar === 'undefined') ? 
+							<Sidebar app={this} navigation={this.Navigation} display="lg"/> : 
+							<Sidebar app={this} navigation={this.Navigation} display="xs"/>}
```
ateska, line 172: You still want to display Sidebar when `this.props.hasSidebar` is `False`.

Pe5h4: Yes. When False, it will display Sidebar and items in it only if there are any items added to the Header, otherwise it will not display the sidebar. It basically resolves the Header issue with collapsing the Header items / components into a toggler button when you are on smaller screens. 
Let say, that Header contains e.g. 2 buttons, <button1/> and <button2/>. If you want to use them and collapse them, when the screen is diminished, you can add them as items into the Navigation of the Sidebar, even though the Sidebar is "disabled" by False. Then, when the screen change the size (e.g. when you diminish the browser screen or you are on your mobile phone), the items from the Header disappear and SidebarToggler button will appear with the <button1/> and <button2/> options. So you can access their functions on smaller screens.
If Sidebar is set to False and no items / components are added to the Header, Sidebar will not show up at all.

src/containers/Application.js
Pe5h4: Changed in last commits

PR #38, opened 2020-07-31
Fixes basePath handling.

awichera:


src/containers/Application.js
```diff
@@ -93,10 +93,19 @@ class Application extends Component {
...
+					API_URL: publicUrl + '/api/',
+				});
+			}
+			else {
+				this.Config.addDefaults({
+					API_URL: window.location.protocol + '//' + window.location.host + '/api/',
+				});
+			}
 		}
+		console.log("passed");
```
ateska, line 108: No console.log() in the merge.

src/containers/Application.js
```diff
@@ -93,10 +93,19 @@ class Application extends Component {
...
-			});
+			const publicUrl = __CONFIG__["publicUrl"];
+			if (publicUrl != ""){
+				this.Config.addDefaults({
+					API_URL: publicUrl + '/api/',
+				});
+			}
+			else {
+				this.Config.addDefaults({
+					API_URL: window.location.protocol + '//' + window.location.host + '/api/',
```
ateska, line 104: Well, I expected that this will go into basepath, not here.

ateska: The logic of "API_URL" should be:
1) It is specified in the config, so it will be used
2) when it is not specified, the base path is taken and "/api" is added (no `/` at the end).

"publicUrl" will go away eventually, as discussed, don't reintroduce it.

ateska:
Closing since this is likely stalled.

PR #49, opened 2020-09-21
Feature/i18n update

Pe5h4:
Changes made as a result of review on branch https://github.com/TeskaLabs/asab-webui/tree/feature/i18n . It concerns the demo application and the result is facilitation for user when adding translations to its project. The `locales` folder was added into the demo application as an example to avoid misleading user when implementing translations.

### DIFF
--- a/demo/public/locales/en-US/translation.json
+++ b/demo/public/locales/en-US/translation.json
@@ -1,0 +1,4 @@
+{
+	"Welcome to React": "Vitejte v Reactu",
+	"Good day": "Dobry den"
+}
--- a/demo/public/locales/en-US/translation.json
+++ b/demo/public/locales/en-US/translation.json
@@ -1,4 +1,4 @@
 {
-	"Welcome to React": "Vitejte v Reactu",
-	"Good day": "Dobry den"
+	"Welcome to React": "Hello my friend",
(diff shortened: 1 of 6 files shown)


demo/public/locales/en-US/translation.json
```diff
@@ -0,0 +1,4 @@
+{
+	"Welcome to React": "Vitejte v Reactu",
```
ateska, line 2: This is not english (and I'm not sure if we need `en-US` folder

demo/public/locales/en/translation.json
```diff
@@ -0,0 +1,4 @@
+{
+	"Welcome to React": "Vitejte v Reactu",
```
ateska, line 2: Must be english.

ateska (changes requested):
Small changes in the text, then mergeable.

demo/public/locales/en-US/translation.json
Pe5h4: Concerning `en-US` folder - it does not need to be there, but if it is not added, then we have a warning with failed localizations ```something went wrong loading 
Array [ "failed parsing /locales/en-US/translation.json to json" ]```

ateska (approved):

PR #50, opened 2020-09-21
Create Tools dashboard

Pe5h4:
This PR should create Tools dashboard, where external links should be displayed as icons/images with external links. Configuration should be passed from site or it can be static. Since it was implemented as a Module, it can be optionaly used in the applications build on top of ASAB-WebUI.


doc/tools.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+
+## Setup in config file
+
+To load the configuration, external url's and paths of the image locations must be set.
+
+```
+{
+	"TOOLS":{
```
ateska, line 31: "tools":  ...

And the content should be array (name is the attribute within).

ateska (changes requested):

ateska:
I would be cool to see a screenshot too :-)

Pe5h4:
1-tools-link-on-hover
2-tools-module

ateska (approved):

ateska:
Try to put the icon into the <Button>

PR #63, opened 2020-11-23
Fix/languagedropdown display bug

Pe5h4:
This PR fix language dropdown bug, when the project was build with `yarn build` command. Previously, the dropdown was not available on small screens (cellphones). The reason was, that when build by `yarn build` command, the name of the dropdown has been changed from `LanguageDropdown` to `jo` and thus has not been catched by the condition. The solution was just to add the name to the component props of the dropdown. Also default i18n keySeparator of the project has been mentioned in the documentation.

### DIFF
--- a/doc/i18n.md
+++ b/doc/i18n.md
@@ -43,6 +43,8 @@ const ConfigDefaults = {
 
 ## Use in React Hooks
 
+Default key separator is set to `|`.
+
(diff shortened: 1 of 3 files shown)


Pe5h4:
@ateska

ateska (approved):

PR #64, opened 2020-12-07
Add the logic to handle small numbers (between 0 and 1)

wuujcik:
Add the logic to handle small numbers (between 0 and 1). Add basic handling of negative numbers (basically display they as they are instead of "undefined"

### DIFF
--- a/src/containers/Humanize.js
+++ b/src/containers/Humanize.js
@@ -1,7 +1,21 @@
 import React, { Component } from 'react'
 
 /*
-Diplays values in human readable form.
+Diplays values in human readable form, like:
+0.000001 => 1 µ
+0.00001 => 10 µ
+0.0001 => 100 µ
+0.001 => 1 m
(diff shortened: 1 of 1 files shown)


Pe5h4 (approved):

src/containers/Humanize.js
```diff
@@ -42,13 +56,30 @@ export function Humanize(props) {
 
 function formatInput(value, base, decimals, displaySizes) {
 	if (value === 0) return '0';
+	if (value < 0) return value;
```
ateska, line 59: Well, you should support also negative values ...

ateska (changes requested):
Negative values should be supported as well.

src/containers/Humanize.js
wuujcik: @ateska  updated to support also negative values

ateska (approved):

PR #65, opened 2020-12-17
Set title and site_title in branding and display it in the page title

wuujcik:
Added a config for parent_brand_image that will be displayed in header
image


ateska:
I don't recall why we need to have two branding icons ... can you remind me, please/

wuujcik:
> I don't recall why we need to have two branding icons ... can you remind me, please?

@ateska so that we can display whatever Customer wants and then our logo, like "Customer's logo | Logman.io" (task INDG-9). I understood it should be part of this repo, but if you think it should be somewhere else, let me know pls

wuujcik:
@ateska updated as per conversation to use our brand name in title, not header

demo/src/index.js
```diff
@@ -8,6 +8,7 @@ import './index.scss'
 
 // Configuration
 let ConfigDefaults = {
+	site_title: "TeskaLabs",
```
ateska, line 11: The default should be empty.

ateska (changes requested):
One smaller remark, otherwise mergeable from my perspective.

demo/src/index.js
wuujcik: actually I think it should be `site_title: "ASAB App"` and `title` should be empty (since the title has format `site_title | title`), correct?

src/containers/Application.js
```diff
@@ -297,8 +299,14 @@ it is accessible by the sidebar toggler button.
...
 				<div className="app">
-					<AlertsComponent app={this}/>
-					<SplashScreen app={this}/>
+					<AlertsComponent app={this} />
+					<SplashScreen app={this} />
+					{this.Config.get('site_title') != null && this.Config.get('site_title') != undefined ?
+						<Helmet>
+							<title>{this.Config.get('title') ? this.Config.get('site_title') + " | " + this.Config.get('title') : this.Config.get('site_title')}</title>
+						</Helmet>
+						: { null }
```
Pe5h4, line 308: @wuujcik Not sure about the `null` wrapped in curly brackets here and on line 325 https://github.com/TeskaLabs/asab-webui/pull/65/files#diff-31d3784ef11656fbccbce9b6b554c572b43dfad75c897ef7fa510ba7a443735fR325. I would suggest to remove the `null` brackets, otherwise it can cause crash of the app.

demo/src/index.js
ateska: Format is to be either:

`title` or `site_title | title`

src/containers/Header.js
```diff
@@ -43,7 +43,7 @@ export function Header(props) {
 					width: 30,
 					height: 30,
 				}}
-			/>
+			/> |
```
Pe5h4, line 46: @wuujcik You likely forgot to remove the separator here.

Pe5h4 (changes requested):

ateska (approved):

Pe5h4 (approved):

PR #67, opened 2021-01-11
Implement ASAB config module

Pe5h4:


src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,222 @@
...
+// TODO: Different types of ConfigItem to cover formats such as "number", "boolean", checkbox, radiobox
+function ConfigItem(props) {
+	let myid = '['+props.sectionname + "] " + props.itemname;
+	return (
+		<FormGroup>
+			<Label for={myid}>
+				{props.item['title']}
+			</Label>
+			<ConfigItemFormats
+				format={props.item['type']}
```
Pe5h4, line 147: @ateska Hope this is not much of stupid question, but will be input formats taken from `type` property or will be there any format property in the schema to distinguish, which format to take into consideration? I updated the component `ConfigItem` with `ConfigItemFormats` and this require to pass to props a format type, to select the proper one. I set it to the `type` property for now, but it can be easily changed. So my question is, if I should substract it from `type` property, or is it defined somewhere else in the schema?

ateska: Well, I rather expect various specializations of `ConfigItem` .... eg `<EmailConfigItem>` or `<CheckboxConfigItem>`.

The type should be based on "some" property in the schema ... and this shoud be so far relatively easy mapping (e.g. by default, it is always `ConfigItem`, if specific format is requested, then you use more specific ConfigItem`.

ateska (changes requested):

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,216 @@
...
+	return (
+		<Card style={{marginBottom: "0.25em"}}>
+			<CardHeader tag="h5" onClick={toggle}>
+				{props.section['title']}
+			</CardHeader>
+			<Collapse isOpen={isOpen}>
+				<CardBody>
+					{Object.keys(props.section.properties).map((item_name, idx) =>
+						// Decide what type of config item to render based on format
+						// TODO: `format` names should be consulted with BE team
```
ateska, line 135: It is important to align this with a possibilities of the JSON schema.
But we can start with very "weak" validation, so that all of fields are single-line text inputs.
That will allow this PR to be merged.

ateska (changes requested):

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,216 @@
...
+				console.log(error); // log the error to the browser's console
+				App.addAlert("warning", t(`Unable to get ${typeId} data: `, { error: error.toString() }));
+			});
+		}
+	}
+
+	useEffect(() => {
+
+		async function fetchValues() {
+			// let config_response = await axios.get("/config/new_type/my_config")
```
ateska, line 61: /config/new_type/my_config?format=json

src/modules/asab-config/ConfigEditor.js
```diff
@@ -28,6 +28,7 @@ export default function ConfigEditor(props) {
 
 	const [ typeId, setTypeId ] = useState(props.match.params.type_id);
 	const [ type, setType ] = useState(undefined);
+	const [ adHocValue, setAdHocValue ] = useState({});
```
ateska, line 31: `adHocSections`, I guess ... the default should be `[]` ... otherwise I don't get it.

ateska (changes requested):
The concept of AdHoc sections needs to be a bit extended.

ateska (changes requested):

Pe5h4:
@ateska I have refactored the asab-config structure as you suggested to:

`ConfigContainer` where TreeView and Configeditor components are rendered

`TreeViewComponent` where data for tree view are fetched and rendered (I must have done several calls to asab-config service to obtain all the types and files, because there is no single API call to obtain the structure)

`ConfigEditor` component where schema and config data are rendered. I have updated the the editor and used only one useEffect function to obtain all the data needed. I have also created 2 AdHoc parts > one is for AdHocSections and the second is for AdHocValues (which are displayed in the ConfigSection, as you can see in oneof the attached gifs)

`ConfigFormatItems` from where various ConfigItems with various formats are taken to render.

In the next iteration, I would implement routering for types and configs (now the url is constant) and data fetching for treeview (it might be cool to create api endpoint for that, or idk)

Please check it and let me know, what should I refactor.

Asab-config
asab-config

Asab-config with mocked data to simulate AdHocValues in sections
asab-config_adhoc

ateska:
Good good.
Remarks:

1) The navigation in the tree view should change the URL - so that I can share a link to particular configuration
2) The name of the actual config should be a simple "type/name" form ... not two lines with that icon.

src/modules/asab-config/ConfigContainer.js
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	return (
+		<Container fluid className="animated fadeIn flex mt-0 pr-0 pl-0 pt-0 library-container">
+			<Row className="library-row">
+				<Col xs="2" sm="2" className="pr-0 bcg-column">
+					<TreeViewComponent
+						axios={Axios}
+						app={App}
+						configType={configType}
```
ateska, line 42: You don't need to pass `axios` nor `app` ... `app` is in `props` automatically, Axios should be local.

Plus it makes component more universal.

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,364 @@
...
+	useEffect(() => {
+		initialLoad();
+	},[configName]);
+
+
+	const initialLoad = async () => {
+		let values = undefined;
+		let type = undefined;
+
+		if (configType != "") {
```
ateska, line 53: Nope, no `if` like that - it has to be present, you don't need to check it.

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,364 @@
...
+			}
+			catch {
+				App.addAlert("warning", `Unable to get ${configType} data`);
+				return;
+				// TODO: Prepared for i18n
+				// App.addAlert("warning", t(`Unable to get ${typeId} data: `, { error: error.toString() }));
+			}
+		}
+
+		if (configName != "") {
```
ateska, line 68: Dtto.

ateska (changes requested):

src/modules/asab-config/ConfigContainer.js
Pe5h4: Well, TreeViewComponent is a component of the ConfigContainer - we need to pass the app into the props of that component, it is not passed automatically here.

Pe5h4:
@ateska Added navigation by url params, updated the name of the actual config, updated alert messages.
Here it is how it looks like with the navigation:
asab-config_urlparams

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,370 @@
...
+export function ConfigEditor(props) {
+
+	const [ typeData, setTypeData ] = useState(undefined);
+	const [ adHocValues, setAdHocValues ] = useState({});
+	const [ adHocSections, setAdHocSections ] = useState({});
+	const { register, handleSubmit, setValue, getValues, errors, reset } = useForm();
+
+	let App = props.app;
+
+	let urlParams = new URLSearchParams(App.props.location.search);
```
ateska, line 38: Nope, the correct approach is:

```
const configType = props.configType
const configName = props.configName
```

The rest is misconception.

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,370 @@
...
+	let url = services?.asabconfig ? services.asabconfig : 'asab-config';
+	const Axios = App.axiosCreate(url);
+
+
+	// Load Schema type - it is triggered on every configName change
+	useEffect(() => {
+		if (configType && configName) {
+			initialLoad();
+		}
+	},[configName]);
```
ateska, line 56: Nope, this is really only initial load, so effect is on `[]`

src/modules/asab-config/ConfigEditor.js
```diff
@@ -0,0 +1,370 @@
...
+
+	// Retrieve the ASAB_CONFIG_URL from config file
+	let services = App.Config.get('SERVICES');
+	let url = services?.asabconfig ? services.asabconfig : 'asab-config';
+	const Axios = App.axiosCreate(url);
+
+
+	// Load Schema type - it is triggered on every configName change
+	useEffect(() => {
+		if (configType && configName) {
```
ateska, line 53: No if ... please, config_type and config_name HAS TO be present ... if you encounter situations, when they are not, then you have incorrect design.

ateska (changes requested):

src/modules/asab-config/ConfigContainer.js
```diff
@@ -0,0 +1,52 @@
...
+} from "reactstrap";
+
+import { TreeViewComponent } from "./TreeViewComponent";
+import { ConfigEditor } from "./ConfigEditor";
+
+export default function ConfigContainer(props) {
+
+	let App = props.app;
+
+	const [ configType, setConfigType ] = useState(undefined);
```
ateska, line 22: This is not a state, it should came from a router, please.

src/modules/asab-config/index.js
```diff
@@ -0,0 +1,29 @@
...
+export default class ConfigModule extends Module {
+	constructor(app, name) {
+		super(app, "ASABConfigModule");
+		this.App = app;
+		this.Config = app.config;
+		this.Navigation = app.Navigation;
+		this.Router = app.Router;
+
+		this.Router.addRoute({
+			path: "/config",
```
ateska, line 14: `/config/:config_type/:config_name` is a must.

ateska (changes requested):

src/modules/asab-config/index.js
```diff
@@ -25,11 +25,10 @@ export default class ConfigModule extends Module {
 			component: ConfigContainer,
 		});
 
-		// TODO: add routes for types and configs
 
 		this.Navigation.addItem({
 			name: "Configuration",
-			url: "/config/*/*",
+			url: "/config/type/name",
```
ateska, line 31: Better 'empty state' variables are recommended here ... such a `/config/$/$` so that you prevent a collision with reasonably named files.

src/modules/asab-config/ConfigContainer.js
```diff
@@ -33,13 +33,23 @@ export default function ConfigContainer(props) {
...
-						configName={configName}
-					/>
+					{configType != 'type' && configName != 'name' ?
+						<ConfigEditor
+							app={App}
+							configType={configType}
+							configName={configName}
+						/>
+					:
+						<React.Fragment>
```
ateska, line 43: This is called "empty state" -> https://www.toptal.com/designers/ux/empty-state-ux-design

Very interesting topic, btw.


Also, you don't need to use "React.Fragment" when you have only one element in that.

ateska (changes requested):

Pe5h4:
@ateska Updated visualization of asab-config. The `LogMan.io` image is generic and configurable. It depends on `brand_title`. I think there is a space for those informational screens there and its styling. Should we focus on those in the next iteration or should I do it in this PR?

asab-config_screens

ateska (approved):

ateska:
Looks mergeable to me.

PR #70, opened 2021-01-13
Implement/authorisation on tenants

Pe5h4:


Pe5h4:
@ateska I have implemented authorization of a user by assigned tenants. The user needs to have assigned tenants to get to the application. If not, or if user's tenants do not match with those in the database, it will not allow the user to proceed. User can use only those tenants, which are assigned to him. 

It works, but I would like to have some feedback. I am not sure if I have correctly implemented obtaining of the auth token as seen below, because in console, I am getting typeerror, that `e.OAuthToken is null` which is probably caused by the fact, that it is not in the session storage by the time when the first attempt to obtain it appears. Since I need to build the app, it is difficult for me to distinguish, from where the error particurarly comes from 
Selection_122

. Would it be better to obtain it from the `auth` module?
https://github.com/TeskaLabs/asab-webui/blob/b673ceda12700c1d2b3fbef8b02e352418974041/src/modules/tenant/service.js#L10
Also, should I refactor the files I have changed with the react hooks?

authorization2

src/modules/tenant/api.js
```diff
@@ -1,9 +1,16 @@
...
-    get_tenants() {
-        return this.Axios.get('/tenant')
-    }
+	get_tenants() {
+		return this.Axios.get('/tenant')
+	}
+
+	get_user_access(tenant, access_token) {
+		return this.Axios.get(
+			"/rbac/" + tenant + "/tenant:access",
```
ateska, line 12: This needs to be generalized into auth module as `verify_access([tenant], [resource])` - resource is "tenant:access".

The access token should be handled internally, not to be exposed to the application.

Tenant module may provide this convenience method to `verify_access_to_tenant()`.

src/modules/tenant/service.js
```diff
@@ -7,6 +7,7 @@ export default class TenantService extends Service {
 	constructor(app, name="TenantService") {
 		super(app, name);
 		this.API = new Api(this.App.axiosCreate(''));
+		this.OAuthToken = JSON.parse(sessionStorage.getItem('SeaCatOAuth2Token'));
```
ateska, line 10: Nope, this has to stay within the auth module.

ateska (changes requested):
Good idea, wrong place of implementation.

Pe5h4:
@ateska I have refactored the code and implemented the authorization by tenants to auth module. There is still WIP on API of auth modules, this I will do on monday. I also updated tenant dropdown, that it shows only tenants, that are authorized for particular user and application ( I used redux for this case ).  The access token for `rbac` endpoint is now obtained directly in auth module. I also used the `get_tenants` api call from `tenant module` and use it in `auth module`, because there is a need to obtain the list of tenants of the application to validate tenants assigned to the user.
authorization3

Pe5h4:
@ateska So as I wrote, I have updated the API calls, you can check them here: https://github.com/TeskaLabs/asab-webui/blob/9620fd8144aa84e51ec4a6166f97773a0932e1e6/src/modules/auth/api.js

I have also updated the authorisation flow in here: https://github.com/TeskaLabs/asab-webui/blob/9620fd8144aa84e51ec4a6166f97773a0932e1e6/src/modules/auth/index.js 
, so it should be smoother when loading the application and checking for authorized tenants.

Here is the visual example:
authorization4

src/modules/auth/api.js
```diff
@@ -10,26 +10,49 @@ export class SeaCatAuthApi {
...
+			subpaths: {oidc: '/openidconnect', rbac: '/rbac'},
 			...
 	*/
 
 	constructor(config) {
-		
-		this.BaseURL = config.get('OIDC_URL');
+
+		this.BaseURL = config.get('BASE_URL');
+		this.microservice = config.get('microservice');
```
ateska, line 22: `this.Microservice` ... please be consistent - applies to other member variables too.

src/modules/auth/api.js
```diff
@@ -10,26 +10,49 @@ export class SeaCatAuthApi {
...
+		this.oidc = this.subpaths.oidc ? this.subpaths.oidc : '/openidconnect'; // Openidconnect
+		this.rbac = this.subpaths.rbac ? this.subpaths.rbac : '/rbac'; // rbac
+
 		this.Axios = axios.create({
 			timeout: 10000,
-			baseURL: this.BaseURL,
+			baseURL: this.BaseURL + this.microservice + this.oidc,
 		});
 
+		this.AxiosTenants = axios.create({baseURL: this.microservice});
```
ateska, line 54: Nope. Only one Axios please, with `this.BaseURL` only ... `this.microservice` should go away, there is no reason for it.

`this.OIDC` should be appended on places, where OIDC call is made.

src/modules/auth/index.js
```diff
@@ -42,23 +44,28 @@ export default class AuthModule extends Module {
...
 				// User info not found - go to login
 				sessionStorage.removeItem('SeaCatOAuth2Token');
 				let force_login_prompt = true;
 
-				this.Api.login(this.RedirectURL, force_login_prompt);
+				this.SeaCatAuthApi.login(this.RedirectURL, force_login_prompt);
+				return;
+			}
+			// Authorization of the user based on rbac
+			let userAuthorized = await this._isUserAuthorized();
```
ateska, line 58: I guess this has to be configurable - we have applications with authentication but without autorization.

src/modules/auth/index.js
```diff
@@ -5,20 +5,22 @@ import reducer from './reducer';
...
 
+export default class AuthModule extends Module {
 
 	constructor(app, name){
 		super(app, "AuthModule");
 
 		this.OAuthToken = JSON.parse(sessionStorage.getItem('SeaCatOAuth2Token'));
 		this.UserInfo = null;
-		this.Api = new SeaCatAuthApi(app.Config);
+		this.SeaCatAuthApi = new SeaCatAuthApi(app.Config);
```
ateska, line 16: Nope, it has to still be `Api` since other providers such as Google OAuth could be used eventually.

src/modules/auth/reducer.js
```diff
@@ -2,6 +2,8 @@ import { types } from './actions'
 
 const initialState = {
 	userinfo: null,
+	allowedTenants: null,
+	currentAllowedTenant: null,
```
ateska, line 6: Just `tenants` ... don't know what is a difference among `allowedTenants` and `currentAllowedTenant` ... and there should not be any.

src/modules/auth/reducer.js
```diff
@@ -2,6 +2,8 @@ import { types } from './actions'
 
 const initialState = {
 	userinfo: null,
```
ateska, line 4: Another remark here is that 'tenants' for given credentials should be part of `userinfo` call. No dedicated API call should be necessary for obtaining 'current' tenants, roles, etc..

If SeaCat Auth don't provide this now, please communicate with PLUM team.

ateska (changes requested):

src/modules/auth/api.js
Pe5h4: @ateska Well, then we need to figure out, how to proceed when the `oidc` BaseUrl differs from BaseUrl of the application itself. It can happen e.g. in SeaCat WebUI, when the BaseUrl for authentication is related to SeaCat Auth WebUI (e.g. auth.teskalabs.com), but for API calls after authentication, there is a need to use BaseUrl related to SeaCat WebUI (e.g. admin.teskalabs.com).

src/modules/auth/reducer.js
Pe5h4: @ateska Well, I have renamed it in the last commit to `tenants` and `current` - to give you a small exmplanation, the `tenants` are all tenants, which are authorized for the user to be used in the application (this will not provide you `userinfo`, but it is exctracted from `rbac`). The `current` is the current tenant, which is taken from the url parameters (if defined) or it is the first authorized tenant from the list of authorized tenants. This is handy, when user will log out from the application and in the url will stay the tenant parameter. But when e.g. another user without the authorization for this particular tenant will try to log in, it will not let him in or it will redirect to the part of the application, for which the user has authorized access (by another tenant).

src/modules/auth/reducer.js
```diff
@@ -17,8 +17,8 @@ export default function reducer(state = initialState, action) {
 		case types.AUTH_TENANTS:
 			return {
 				...state,
-				allowedTenants: action.payload,
-				currentAllowedTenant: action.currentAllowed
+				tenants: action.payload,
+				current: action.current
```
ateska, line 21: Ok, I see ... so the name is `activeTenant` ...

ateska (changes requested):

src/modules/auth/reducer.js
ateska: Ok, as commented elsewhere, it is then `activeTenant` ...

Pe5h4:
@ateska I have made some changes based on your comments, but I have also some questions on you:

1] When using Authorization, should TenantDropdown show only tenants, which has been authorized by the particular application, or all the tenants available for the user? What I mean by it - let say, that application has 4 tenants (tenant1, tenant2, tenant3, tenant4) and user has 3 tenants (tenant2, tenant4, tenant666), so only `tenant2` and `tenant4` are authorized for the user and thus user has an access to the application through tenant2 or tenant4. Now it is constructed that way, that TenantDropdown displays for this case just `tenant2` and `tenant4`. Is it ok like that, or should the dropdown display also `tenant666` and when user will try to click on it, it should display warning, that user is not authorized to use this tenant in this application and redirect the user back to one of the authorized tenants ? 

2] I have implemented logout with timeout for non-authorized users, that user can read the message that he/she is not authorized, before it will redirect him/her back to login. (It is because it does not seems right, when there is just splashscreen with the message that user is not authorized. People can be confused.). So my quiestion is, should I leave the logout in Authorization, or should I remove it? https://github.com/TeskaLabs/asab-webui/pull/70/commits/8afbf971ff17c28af3da78a6c961567c3d1561ea#diff-7e330056a24d48cd7e49039fad7cda505b858f305494a462e623ecc43cc51b0fR64

3] As your comment here: https://github.com/TeskaLabs/asab-webui/pull/70/files/9620fd8144aa84e51ec4a6166f97773a0932e1e6#diff-3e8a0c0c99df01fe82eb6d8165ad91a7af628f2d35ab393c3ecc65dbf275eb6eR54
How could we proceed when the `oidc` BaseUrl differs from BaseUrl of the application itself? As I wrote in that response, it can happen e.g. in SeaCat WebUI, when the BaseUrl for authentication is related to SeaCat Auth WebUI (e.g. auth.teskalabs.com), but for API calls after authentication, there is a need to use BaseUrl related to SeaCat WebUI (e.g. admin.teskalabs.com).

src/modules/auth/reducer.js
Pe5h4: Sure, I can get the tenants for given credentials from `userInfo`, but I dont see the point, because authorization dont need the user tenants. We might need them if we will decide, that in the TenantDropdown will be displayed all the tenants assigned to the user, not only the authorized tenants. But this is what I mentioned in `1]` question on you on my latest comment.

Anyway, in this case, the reducer pass the payload of authorized tenants, which has been allowed to use within the application (this is substracted from the `rbac` endpoint, but this endpoint require the tenants of the application, not the user's). This payload is passed as a props to the TenantDropdown and then displayed. So only tenants for which user has authorization are displayed.

So, there is no dedicated API call for obtaining `current` tenants, roles, etc... But there is an API call for obtaining all the application's tenants, because they are needed for authorization procedure, which I have mentioned above.

Pe5h4:
@ateska I have a question on you regarding the CORS policy on `rbac` endpoint. I am not able to obtain the result from this endpoint thanks to the CORS and I wonder, how can I pass it? The reason of those cors policy errors are `preflights` in our case, but I have no clue, how to fix it. I tried to allow origin `Access-Control-Allow-Origin: '*'`, allow methods `Access-Control-Allow-Methods: GET, OPTIONS`, set cross domain to true `crossdomain: true` , but nothing helped.
https://github.com/TeskaLabs/asab-webui/pull/70/files#diff-3e8a0c0c99df01fe82eb6d8165ad91a7af628f2d35ab393c3ecc65dbf275eb6eR111

I am not able to localize, what is wrong then and what needs to be changed. I use the BaseURL for every `oidc` call and also for obtaining the tenants of the application. Everywhere it works, but only when trying to use it with `rbac` , I will encounter the CORS. May it be because of the `accessToken` used for verifying the user in `rbac` ? Anyway, I made some printscreens and gif video

Selection_125
Got `405` error when trying to use the `rbac` endpoint
Selection_124
Network flow before it crash on cors
authorization-cors2

ateska:
Here is my take on the approach to API URL composition:

# Default URL composition

{
	base_url: "http://localhost:3000",
	api_path: "/api",
	services: {
		oidc: "/openidconnect",
		rbac: "/rbac",
		...
	}
}

URL = base_url + api_path + services.{name}

# Absolute URL option

{
	services: {
		oidc: "https://example.com/openidconnect"
	}
}

If URL starts with protocol (http:// or https://), then use that URL as it is

URL = services.{name}

# app.axiosCreate

Introduce a new method which obsoletes the `app.axiosCreate` (eventually)

`app.axiosCreateService({name}, props)`

or explore the alternative immediate refactoring to:

`axiosCreate({name}, props)`

`{name}` is not `path` as today but a reference to "services" in the config.

README.md
```diff
@@ -30,12 +30,12 @@ It is designed primarily for web-based user interfaces for ASAB microservices bu
 TODO ...
 
 ```
-[ASAB UI App] -- [ HTTP proxy (API_URL) ] -- [ actual microservices ]
+[ASAB UI App] -- [ HTTP proxy (ASAB_MICROSERVICE) ] -- [ actual microservices ]
```
ateska, line 33: `ASAB_MICROSERVICE` this term should not be used - it is a `SERVICE_URL` ... but look at the URL composition proposal first.

doc/authorization.md
```diff
@@ -0,0 +1,54 @@
...
+const modules = [];
+
+...
+
+import TenantModule from 'asab-webui/modules/tenant';
+modules.push(TenantModule);
+
+let ConfigDefaults = {
+	ASAB_Authorization: true,
+	ASAB_RBAC_Resource: "tenant:access",
```
ateska, line 44: This is not a good name, don't prefix with "ASAB" ... use section with name 'authorization'

ateska (changes requested):

Pe5h4:
@ateska I made changes according to your proposal. There is still one thing that needs to be solved - CORS policy.

I am still getting CORS errors when trying to obtain results from `rbac` endpoint. But it is not all - if one would like to make some change e.g. in adminpanel of the application (e.g. creating a new user), it will fail again on cors policy. Honestly, I have no clue, how to resolve it when we are going to use `BASE_URL` for axios create. https://github.com/TeskaLabs/asab-webui/pull/70/files#diff-31d3784ef11656fbccbce9b6b554c572b43dfad75c897ef7fa510ba7a443735fR166

To give you better overview on it, here are errors I am getting and you can also see pictures and gif with the errors in my previous comment https://github.com/TeskaLabs/asab-webui/pull/70#issuecomment-766917063

Trying to fetch the response from `rbac`
```
Access to XMLHttpRequest at 'http://auth.test.localhost/api/rbac/tenant1/tenant:access' from origin 'http://adminpanel.test.localhost' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: It does not have HTTP ok status.
```

Creating a new credentials
```
Access to XMLHttpRequest at 'http://auth.test.localhost/api/credentials/ext' from origin 'http://adminpanel.test.localhost' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: It does not have HTTP ok status.
```

config/defaults.js
```diff
@@ -5,7 +5,7 @@ const CWD = process.cwd();
 module.exports = {
 	app: {
 		"publicUrl": "",	// Overriden by --public-url [PUBLIC_URL] command line argument
-		// API_URL: "/api/",
+		// ASAB_MICROSERVICE: "/api/",
```
ateska, line 8: Wrong name.

demo/conf/config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
 	app: {
-		API_URL: 'api/',
+		ASAB_MICROSERVICE: 'api/',
```
ateska, line 3: Wrong name.

ateska (changes requested):
Small remarks, then mergeable.

Pe5h4:
@ateska I made fixes in naming, but I have a 2 questions:

1] Do we need `publicUrl` and `getBaseName` script at all when we use `HashRouter` ? https://github.com/TeskaLabs/asab-webui/blob/implement/authorisation-on-tenants/src/basename.js#L16
And is `publicUrl` nowadays `BASE_URL` ?

2] Still having CORS issues as mentioned in my comments here: https://github.com/TeskaLabs/asab-webui/pull/70#issuecomment-766917063 and here: https://github.com/TeskaLabs/asab-webui/pull/70#issuecomment-768406131
Is there any way how to solve it? I suppose its nginx issue, but still...

ateska:
1) `publicUrl` and `BASE_URL` ... I guess they are the same ... but I don't know if we can get rid of publicUrl easily. Let's investigate (in another PR, maybe?)

2) getBaseName() is probably not needed when on `HashRouter` - I was not fully committed to `HashRouter` so I kept getBaseName() ... please add a deprecation notice to getBaseName, we will remove that maybe in ~6 months from now

3) When you access different domains (in more complicated deployments), then CORS has to be properly configured in Nginx, I guess we cannot do much with this ...

src/containers/Application.js
```diff
@@ -158,9 +158,22 @@ it is accessible by the sidebar toggler button.
...
 	axiosCreate(path, props) {
+		// Set up URL
+		var URL = undefined;
+		if (this.Config.get('BASE_URL') == undefined) {
+			URL = this.Config.get('API_PATH') + path;
+		} else {
+			URL = this.Config.get('BASE_URL') + this.Config.get('API_PATH') + path;
+		}
+
+		// Check for external OIDC of auth module
```
ateska, line 169: Not OIDC specific.

src/modules/auth/api.js
```diff
@@ -10,31 +10,52 @@ export class SeaCatAuthApi {
...
-			this.BaseURL = config.get('seacat.auth.oidc_url');
+			console.log("Provide config value BASE_URL");
+			this.BaseURL = "";
 		}
-		if (this.BaseURL == null) {
-			console.log("Provide config value seacat.auth.oidc_url");
-			this.BaseURL = "/openidconnect";
+
+		if (this.ApiPath == null) {
+			console.log("Provide config value API_PATH");
```
ateska, line 33: `console.log("Config value API_PATH not provided, using /api");`

src/modules/auth/index.js
```diff
@@ -5,20 +5,22 @@ import reducer from './reducer';
...
 
+export default class AuthModule extends Module {
 
 	constructor(app, name){
 		super(app, "AuthModule");
 
 		this.OAuthToken = JSON.parse(sessionStorage.getItem('SeaCatOAuth2Token'));
 		this.UserInfo = null;
-		this.Api = new SeaCatAuthApi(app.Config);
+		this.Api = new SeaCatAuthApi(app.Config, app);
```
ateska, line 16:
```
this.Api = new SeaCatAuthApi(app);
```

... then

src/modules/auth/api.js
```diff
@@ -10,31 +10,52 @@ export class SeaCatAuthApi {
...
 	*/
 
-	constructor(config) {
-		
-		this.BaseURL = config.get('OIDC_URL');
+	constructor(config, props) {
+
+		this.BaseURL = config.get('BASE_URL');
+		this.ApiPath = config.get('API_PATH');
+		this.Services = config.get('SERVICES');
```
ateska, line 23: You definitively don't need to work with `BASE_URL`, `API_PATH` and `SERVICES` ... just get axios using the service name.

ateska (changes requested):

src/modules/auth/api.js
Pe5h4: @ateska Well, I suppose it is needed at least as local variables in constructor. Because it forms the URL for login:
https://github.com/TeskaLabs/asab-webui/blob/implement/authorisation-on-tenants/src/modules/auth/api.js#L63
and services for axios https://github.com/TeskaLabs/asab-webui/blob/implement/authorisation-on-tenants/src/modules/auth/api.js#L43

ateska: Then you rather need a new method on Application, with the similar/same logic as is in `createAxios()` - you don't want to spill URL logic all over the code base.

That new method `getApiURL(service)` should be also used by `createAxios()`

src/modules/auth/index.js
```diff
@@ -5,20 +5,22 @@ import reducer from './reducer';
...
 		this.UserInfo = null;
-		this.Api = new SeaCatAuthApi(app.Config);
+		this.Api = new SeaCatAuthApi(app);
 		this.RedirectURL = window.location.href;
 		this.MustAuthenticate = true; // Setting this to false means, that we can operate without authenticated user
 
 		app.ReduxService.addReducer("auth", reducer);
 		this.App.addSplashScreenRequestor(this);
+
+		this.Authorization = app.Config.get("Authorization"); // Get Authorization settings from configuration
```
ateska, line 23: `this.Config` will be more to-the-point name.

src/modules/auth/index.js
```diff
@@ -53,6 +54,18 @@ export default class AuthModule extends Module {
...
 			}
+			// Authorization of the user based on rbac
+			if (this.Authorization?.Authorize) {
+				let userAuthorized = await this._isUserAuthorized();
+				if (!userAuthorized) {
+					this.App.addAlert("danger", "You are not authorized to use this application.",  60000);
+					// Logout after some time
+					setTimeout(() => {
+						this.logout();
+					}, 60000);
```
ateska, line 65: This could be a configurable value, btw.

src/modules/auth/index.js
```diff
@@ -122,4 +135,67 @@ export default class AuthModule extends Module {
 		return true;
 	}
 
+
+	async _isUserAuthorized() {
+		let resp = false;
```
ateska, line 140: Well, I'm not sure if you actually do authorization check here.
I see that you download tenants.

What you actually need to do, it to call 'rbac' API endpoint and decide on the result.

Tenants should be downloaded else where - likely not in auth. module (that should to some extent work on non-tenant installation as well). It was also reason for my previous comment, that "assigned" tenants should be part of the "userinfo" call, handled elsewhere.

ateska (changes requested):

src/modules/auth/index.js
Pe5h4: @ateska Well, I understand and correct me if I am wrong, but there is a thing, that from `userinfo` the app will get only the tenants assigned to a user and not all the tenants of the application. So then the `rbac` will always pass and there is nothing to check. Because `rbac` should validate, if the users tenants are authorized for the application, am I right? So I need to check all the application tenants (not the users tenants) and validate them with rbac based on user (the information about which user is taken from `acces token`). So that is why authorization needs to get all the tenants of the application - and this is what (now) you dont get from `userinfo`.

In the section, you have commented - I actually download application tenants and then checking them with `rbac` call. And there I validate, if the user tenants are valid for the application and there I get the result if it is valid and check the authorization. The one of the possibilities is that I move the downloading of applicaiton tenants to the Application.js file and then I will pass it to Auth module to avoid loading tenants in the Auth module.

ateska (changes requested):

src/containers/Application.js
```diff
@@ -157,10 +176,43 @@ it is accessible by the sidebar toggler button.
...
+		let API_URL = this.Config.get('API_URL');
+		// return API_URL + "/" + service;
+		return API_URL + "/" + service_path;
+	}
+
+
+	axiosCreate(service, props) {
+		var service_url = this.getApiURL(service);
+		if (service_url == undefined) {
+			return
```
ateska, line 210: `return undefined;`

src/modules/auth/api.js
```diff
@@ -10,31 +10,39 @@ export class SeaCatAuthApi {
...
-		}
-		if (this.BaseURL == null) {
-			console.log("Provide config value seacat.auth.oidc_url");
-			this.BaseURL = "/openidconnect";
+	constructor(app) {
+
+		this.App = app;
+		this.URL = this.App.Config.get('API_URL');
+
+		let Services = this.App.Config.get('SERVICES');
```
ateska, line 24: Nope, please use `app.getApiURL("oidc");`  etc.

src/modules/auth/api.js
```diff
@@ -10,31 +10,39 @@ export class SeaCatAuthApi {
...
+			Services = {"oidc": "openidconnect", "rbac": "rbac"};
 		}
-		this.Axios = axios.create({
-			timeout: 10000,
-			baseURL: this.BaseURL,
-		});
 
-		const scope = config.get('seacat.auth.scope');
+		this.OidcSubpath = Services.oidc ? Services.oidc : 'openidconnect'; // Openidconnect
+		this.RbacSubpath = Services.rbac ? Services.rbac : 'rbac'; // rbac
```
ateska, line 32: You certainly don't need to do this, use `app.getApiURL()`

src/modules/auth/api.js
```diff
@@ -48,12 +56,18 @@ export class SeaCatAuthApi {
 		if (force_login_prompt === true) {
 			params.append("prompt", "login");
 		}
-		window.location.replace(this.BaseURL + "/authorize?" + params.toString());
+
+		let url = this.URL + this.OidcSubpath;
```
ateska, line 60: `app.getApiURL()`

ateska (changes requested):
Still not there.

src/modules/auth/index.js
```diff
@@ -122,4 +137,28 @@ export default class AuthModule extends Module {
...
+		// TODO: Solve race condition when obtaining tenant from redux store
+		const state = this.Store.getState();
```
Pe5h4, line 144: @ateska Here I have a problem to get the current tenant from the store because of race condition. I cannot await it here, so I tried to dispatch it with promise in tenant reducer, but I have not been successful, so temporarily I solved it with the solution that when `activeTenant` (current) is null, then it will take the current tenant from the parameters of the browser. Do you have a hint, how to solve this issue?

Pe5h4:
@ateska I have implemented remarks taken from our workshop. I have one issue with redux store described here: https://github.com/TeskaLabs/asab-webui/pull/70#discussion_r569568469

Also tenants to the tenant dropdown in the header are taken from api, from which all the application tenants are fetched. So in the next MR I will fix that and will fetch the tenants for tenant dropdown from `userinfo`.

src/modules/auth/index.js
ateska: The promise seems to be a good idea. Let's address this in a new MR.

ateska (approved):

PR #77, opened 2021-02-10
Update tenant dropdown

Pe5h4:
Update tenant dropdown with tenants loaded from userinfo to display just tenants assigned to user. Fallback is tenants list obtained from tenant service.
userinfotenantsdropdown


Pe5h4:
@ateska ^^^

src/modules/tenant/TenantDropdown.js
```diff
@@ -26,8 +26,19 @@ class TenantDropdown extends Component {
 
 		this.TenantService = props.app.locateService("TenantService");
 
-		this.tenants = this.props.tenants;
+		// Check on tenants loaded from userinfo (array of strings) or from tenant service (array of objects)
```
ateska, line 29: There shouldn't be a difference - it there is, then userinfo implementation is wrong.

src/modules/tenant/TenantDropdown.js
```diff
@@ -83,7 +94,7 @@ class TenantLabel extends Component {
 const mapStateToProps = state => {
 	return {
 		current: state.tenant.current,
-		tenants: state.tenant.tenants,
+		tenants: state?.auth?.userinfo.tenants ? state?.auth?.userinfo.tenants : state.tenant.tenants,
```
ateska, line 97: `state.tenant.tenants` should be filled from `state?.auth?.userinfo.tenant` after receiving the response from userinfo.
So this line in incorrect.

ateska (changes requested):

src/modules/tenant/TenantDropdown.js
Pe5h4: @ateska userinfo returns a list of strings `tenants = ["tenant1", "tenant2"]` (this is what you will get from API on `/userinfo`), but tenant service returns (I would say more accurate) list of objects `tenants = [{_id:"tenant1"}, {_id:"tenant2"}]` (this is what you will get from API on `/tenants`). Should I communicate that with plum team?

Also, if tenants for authentication should be taken from userinfo, then there is no need to use `/tenants` endpoint, or is it needed for any purpose?

src/modules/tenant/TenantDropdown.js
Pe5h4: @ateska So should I use userinfo API call in tenant module? And since it is used in auth module, would it make sense to move this call to Application.js and dispatch the response from there (to get userinfo in tenant & auth module without separate calls)?

src/modules/tenant/TenantDropdown.js
ateska: After an investigation, `tenants = ["tenant1", "tenant2"]` is the only correct form.

ateska (changes requested):

Pe5h4:
@ateska I am writing you here with the ask for a little help with tenants in ASAB-WebUI, sorry that it is a bit tldr;. I have updated tenants, and with latest change, tenant list should be obtained from userinfo in `auth` module. First, I was thinking to add a new reducer called `tenant` to auth module, but since `userinfo` is called on every page reload, I decided to add list of tenants and the current tenant into `auth` reducer and dispatch it it accordingly i.e. https://github.com/TeskaLabs/asab-webui/blob/fdf89801a0e73a37d6d13acc755786cfc7f30c5e/src/modules/auth/index.js#L127. 

To this point, it works well, but I had to also think, how to get the tenants, when `auth` module is disabled and thus we cannot obtain informations from `userinfo`. And there comes the trouble. I know that the solution I have made is far from ideal, but to this point, I have no idea, how to solve it better way, since my knowledge is not that deep. So, when there is `auth` module disabled, the tenant list is loaded from api endpoint `/tenant` as it was already and then dispatched as `auth` reducer to the store. To enable this behaviour, I need to find a way, how to let `tenant` module know, that `auth` module is disabled. So I used check on OAuthToken from session store (https://github.com/TeskaLabs/asab-webui/blob/fdf89801a0e73a37d6d13acc755786cfc7f30c5e/src/modules/tenant/index.js#L13) But this is initally null even when `auth` module is enabled, so it will pass and thus there is a glitch between reducers with same name `Reducer with name auth already exists.` Because there is one in `auth` module (when `auth` module is enabled) and one in `tenant` module. If the `auth` module is disabled, this problem disappears, because there will be only one reducer with that name (the one used with tenant module)

So, I dont know, if I explained it here understandable, I hope at least partially. What would you suggest? How to approach tenants, when we cannot obtain them from `userinfo`? Should it be dispatched as an empty list and current tenant would be `null` or just use default (initial) values? But there will still be a thing with the `same name` reducer, when both `auth` and `tenant` module will be enabled. Also how can we check, if `auth` module is disabled (I tried to get that information from `App.Modules` and also from `App.Store` but with not much of success)? Or maybe I just took it from the wrong end.

ateska:
Hmms. I don't think that this is a right way of how to propagate tenants into the app.

I suggest to:

- The tenant service should offer a method `set_tenants()` which will basically perform "TENANT_CHANGE" redux action.

- The auth service - after userinfo response with tenants filled - will locate a tenant service and if successful, then it calls `set_tenants()` with info from userinfo.

... and that should do the trick, ok?

ateska (changes requested):
See my comment.

src/modules/auth/reducer.js
```diff
@@ -2,14 +2,19 @@ import { types } from './actions'
 
 const initialState = {
 	userinfo: null,
+	tenants: [],
+	current: null,
```
ateska, line 6: These (`tenants` and `current`) is part of the tenant service, not auth.

src/modules/tenant/actions.js
```diff
@@ -1,4 +1,3 @@
 export const types = {
-	TENANTS_CHANGED: "asab/TenantService/CHANGED",
-	TENANT_SELECTED: "asab/TenantService/SELECT_TENANT",
+	AUTH_USERINFO: "SeaCat/AUTH_USERINFO",
```
ateska, line 2: auth should stay in the auth service  tenant in the tenant service ... mixing it si 100% wrong way.

src/modules/tenant/index.js
```diff
@@ -10,9 +10,12 @@ export default class TenantModule extends Module {
 		super(app, "TenantModule");
 		this.App = app;
 
-		this.TenantService = new TenantService(app, "TenantService");
+		const OAuthToken = JSON.parse(sessionStorage.getItem('SeaCatOAuth2Token'));
+		if (OAuthToken == null) {
+			this.TenantService = new TenantService(app, "TenantService");
+			this.App.ReduxService.addReducer("auth", reducer);
+		}
```
ateska, line 17: No OAuth in tenant service (unless you have very very strong reason for it).

ateska (changes requested):

Pe5h4:
@ateska I have updated the auth and tenant modules as you suggested.

I have created the set_tenants() method, which I have implemented into TenantService and it is used in `auth` module for dispatching to store tenants obtained from `userinfo`.

When `auth` module is disabled, the WebUI application will not provide any tenants and then default/initial state is used for current tenant (`null`) and for list of tenants (empty list). And this is I suppose expected behavior. The url is then without tenant and there is nothing to display in tenant dropdown, as you can see in the image attached below:

Selection_161

Here is the gif image, how the flow looks like when the tenants are obtained from userinfo

asab-webui-userinfotenants

src/modules/tenant/service.js
```diff
@@ -1,57 +1,43 @@
...
+		// If tenant has not been provided in access URL, pick a first tenant from a list
+		if (tenant_id == null) {
+			tenant_id = payload[0];
+			// ... and refresh (reload) the whole web app
+			window.location.replace('?tenant='+tenant_id+'#/');
+			return;
+		}
+
+		// Find the current tenant in the list and extract its
+		let x = payload.filter((item) => { return item == tenant_id } );
```
ateska, line 29: `x` is a wrong name ... say `current_tenant` (or what is it).

ateska (changes requested):
Last remark, then mergeable.

PR #78, opened 2021-02-25
Update/naming

Pe5h4:
Renaming of getApiURL method and auth service to more legible form. Where needed, the backward compatibility has been preserved.

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -111,32 +111,6 @@ it is accessible by the sidebar toggler button.
 		this.Store = createStore(combineReducers(this.ReduxService.Reducers), composeEnhancers(applyMiddleware()));
 
 		this.ConfigService.addDefaults(props.configdefaults);
-		
-		// Set API PATH, if not configured
-		if (this.Config.get('API_PATH') == undefined && this.Config.get('BASE_URL') == undefined) {
-			this.ConfigService.addDefaults({
-				API_PATH: window.location.protocol + '//' + window.location.host + '/api',
-			});
-			console.log("Config value API_PATH not provided, using \"api\"");
-		} else if (this.Config.get('API_PATH') == undefined && this.Config.get('BASE_URL') != undefined) {
-			this.ConfigService.addDefaults({
-				API_PATH: 'api',
-			});
-			console.log("Config value API_PATH not provided, using \"api\"");
-		}
-
-		// Set BASE_URL and API_URL
-		if (this.Config.get('BASE_URL') == undefined) {
-			this.ConfigService.addDefaults({
-				API_URL: this.Config.get('API_PATH'),
-				BASE_URL: "",
-			});
-			console.log("Config value BASE_URL not provided, using \"\" ");
-		} else {
-			this.ConfigService.addDefaults({
-				API_URL: this.Config.get('BASE_URL') + "/" + this.Config.get('API_PATH'),
-			});
-		}
 
 		this.Config.dispatch(this.Store);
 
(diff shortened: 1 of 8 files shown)


src/modules/auth/header.js
```diff
@@ -11,12 +11,15 @@ import {
 function HeaderComponent(props) {
 
 	const Config = props.app.Config;
```
ateska, line 13: You want to do:

```
let user_auth_url = props.app.getServiceURL('auth_webui')
```

... and only if this fails, you handle backward compatibility.
Always try to improve the code when you visit it, it pays of greatly.

ateska (changes requested):
One remark.

src/modules/auth/header.js
Pe5h4: @ateska Well, this will not fail when `auth_webui` will not be defined in SERVICES. If `auth_webui` is not defined in SERVICES, then the `user_auth_url` by your proposal will look like this

```
user_auth_url = 'https://adminpanel.url.com/api/auth_webui'
```

And thus it will not return the desired output of `undefined` and proceed with backward compatibility, because in Application.js it is set like this to allow all the API endpoints (not only those mentioned in SERVICES) and create a service URL from it:

https://github.com/TeskaLabs/asab-webui/blob/093f490ea76cd48f66eba2892532b4c06917dc50/src/containers/Application.js#L208

So if we want to check, if the `auth_webui` is defined/undefined, we need to check it in the configuration and then pass the result to the `getServiceURL(Config.SERVICES.auth_webui)` as it is in proposal in the last commit. https://github.com/TeskaLabs/asab-webui/pull/78/commits/7e49a0d89693a647dde117e0f450f1ca222af9ef

In here, when authPath is undefined, https://github.com/TeskaLabs/asab-webui/blob/7e49a0d89693a647dde117e0f450f1ca222af9ef/src/modules/auth/header.js#L16 , the `user_auth_url` will be also undefined https://github.com/TeskaLabs/asab-webui/blob/7e49a0d89693a647dde117e0f450f1ca222af9ef/src/modules/auth/header.js#L17 and then we can proceed with the backward compatibility

Or then we want to check specifically for `auth_webui` in `getServiceURL` as we do it for `oidc` and `rbac`.

src/modules/auth/header.js
```diff
@@ -10,13 +10,23 @@ import {
...
-	const Config = props.app.Config;
-	let user_auth_url = Config.get('SERVICES')?.auth ? Config.get('SERVICES')?.auth : Config.get('BASE_URL');
-
-	// Provide backward compatibility with USER_AUTH_URL
-	if (Config.get('USER_AUTH_URL')) {
+	const App = props.app;
+	const Config = App.Config;
+	// Get service URL from Config
+	let authPath = Config.get('SERVICES')?.auth_webui;
+	let user_auth_url = App.getServiceURL(authPath);
```
ateska, line 17: `let user_auth_url = App.getServiceURL("auth_webui");`

ateska (changes requested):

src/containers/Application.js
```diff
@@ -176,7 +176,7 @@ it is accessible by the sidebar toggler button.
 	}
 
 
-	getApiURL(service) {
+	getServiceURL(service) {
```
ateska, line 179: getServiceURL(service) {
		let services = this.Config.get('SERVICES');
		if (services == undefined) {
			return undefined;
		}

		service_path = services[service];

		// Handle if service is unknown => return undefined
		if (service_path == undefined || service_path == null) {
			console.log("Service path is undefined");
			return undefined;
		}
		// If service_path is complete URL, then return that URL
		if (service_path.toString().indexOf('http://') !== -1 || service_path.toString().indexOf('https://') !== -1) {
			return service_path;
		}

		let API_URL = this.Config.get('API_URL');
		return API_URL + "/" + service_path;
	}

ateska (changes requested):

Pe5h4:
@ateska Made changes according to the review. I have updated my local nginx settings for SeaCat WebUI with `seacat_auth` and so far it works smooth (according to the changes made in the SeaCat WebUI project).

Have a question about `ConfigService`

How to handle `CONFIG_PATH` with axios, when it is not a service but a path for used in `ConfigService`  here: https://github.com/TeskaLabs/asab-webui/blob/update/naming/src/config/ConfigService.js#L52 ? Should it stay how it is?

ateska (approved):

ateska:
> @ateska Made changes according to the review. I have updated my local nginx settings for SeaCat WebUI with `seacat_auth` and so far it works smooth (according to the changes made in the SeaCat WebUI project).
> 
> Have a question about `ConfigService`
> 
> How to handle `CONFIG_PATH` with axios, when it is not a service but a path for used in `ConfigService` here: https://github.com/TeskaLabs/asab-webui/blob/update/naming/src/config/ConfigService.js#L52 ? Should it stay how it is?

@Pe5h4 - you need to create Axios manually, not using "createAxios" method - since you are "too early" in the init process.

ateska (approved):

PR #80, opened 2021-03-08
Implement/web socket services

Pe5h4:
@ateska I have implemented new modules `getWebSocketURL` and `createWebSocket`. It was tested on LM.io parser builder. Part of this PR is also renaming the `payload` parameter to `tenants_list` in the tenant and auth modules.

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -221,6 +221,66 @@ it is accessible by the sidebar toggler button.
 	}
 
 
+	getWebSocketURL(service, subpath) {
+		// Handle if service is unknown => return undefined
+		if (service == undefined) {
+			console.warn(`WebSocket service is undefined!`);
+			return undefined;
+		}
+		// Handle if subpath is unknown => subpath = ""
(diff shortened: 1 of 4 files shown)


src/containers/Application.js
```diff
@@ -221,6 +221,66 @@ it is accessible by the sidebar toggler button.
 	}
 
 
+	getWebSocketURL(service, subpath) {
+		// Handle if service is unknown => return undefined
```
ateska, line 225: Can we start with:

```
		var service_url = this.getServiceURL(service);
		if (service_url == undefined) {
			this.addAlert('error', "Service URL is undefined, please check service paths passed to axios.");
			return undefined;
		}
```

You repeat a lot of logic here.

ateska (changes requested):

src/containers/Application.js
Pe5h4: You are right, I did the removal of the doubled logic.

ateska (approved):

PR #81, opened 2021-03-09
Initial implementation of http headers passed to axios create

Pe5h4:
Implementation of http headers into axiosCreate method

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -106,7 +106,6 @@ it is accessible by the sidebar toggler button.
 		this.state = {
 			networking: 0, // If more than zero, some networking activity is happening
 			SplashscreenRequestors: 0,
-			AxiosInterceptors: 0,
 		}
 
 		const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
@@ -118,7 +117,6 @@ it is accessible by the sidebar toggler button.
 
 		this.addSplashScreenRequestor(this);
 		this.state.SplashscreenRequestors = this.SplashscreenRequestors.size;
-		this.state.AxiosInterceptors = this.AxiosInterceptors.size;
 
 		var that = this;
 
@@ -206,8 +204,11 @@ it is accessible by the sidebar toggler button.
 			baseURL: service_url,
 		});
 
-		// Add axios interceptors
-		this.addAxiosInterceptors(axios);
+		// Iterate through custom interceptors
+		for (let interceptor of this.AxiosInterceptors.keys()){
+			this.interceptorRequest(axios, interceptor);
+			this.interceptorResponse(axios, interceptor);
+		}
 
 		var that = this;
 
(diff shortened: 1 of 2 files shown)


Pe5h4:
@ateska I did the initial implementation of headers to axiosCreate, as it was requested by Prema regarding the case of passing the Authorization Bearer token in every axios call. 

I would like to ask you, if the approach I took is correct.

1] Should be the default list of the headers in the Application.js ? And how many types should be there? https://github.com/TeskaLabs/asab-webui/blob/229c2540893bd0aaac4f383ab830eb41a706d065/src/containers/Application.js#L86-L89


2] The concept implemented here counts with the header passed within `axiosCreate` method where the API of the service is defined. So I extended it to this point e.g. for Authorization header

```
let MyAPI = App.axiosCreate("myservice")

let My2ndAPI = App.axiosCreate("myservice","Authorization")
```
Now there is only one Authorization header implemented in the HTTPHeaders list of Application.js and it is `Bearer token`. But it can be extended more and have it defined e.g. as

```
let MyAPI = App.axiosCreate("myservice")

let My2ndAPI = App.axiosCreate("myservice",{"Authorization":"Bearer"})

let My3rdAPI = App.axiosCreate("myservice",{"Authorization":"Basic"})

let My4thAPI = App.axiosCreate("myservice",{"Authorization":"Bearer", "Accept":"application/json"})
```

This is in my point of view better, since you can then define multiple items in the header. Then I would refactor the code accordingly.

Let me know please, if I am not much out of the path. Thank you.

ateska:
Comment provided, this is not a correct design.

Pe5h4:
@ateska I refactored the interceptors as you suggested. I also divided interceptor request and response to 2 separate methods `interceptorRequest` and `interceptorResponse`, but I am not sure about this approach

https://github.com/TeskaLabs/asab-webui/blob/ed52f0f01bed107c15c27d03f0bb4c89a010c1da/src/containers/Application.js#L235-L262

I also updated `addAxiosInterceptor` and `removeAxiosInterceptor` methods as you can see here. It just add / remove the interceptor from the `AxiosInterceptors` set. 

https://github.com/TeskaLabs/asab-webui/blob/ed52f0f01bed107c15c27d03f0bb4c89a010c1da/src/containers/Application.js#L265-L272

The interceptors themselves are handled in `axiosCreate` method to apply them in axios calls.

https://github.com/TeskaLabs/asab-webui/blob/ed52f0f01bed107c15c27d03f0bb4c89a010c1da/src/containers/Application.js#L207-L211

Authorization header with bearer access token is created in auth module and passed via `addAxiosInterceptor`

https://github.com/TeskaLabs/asab-webui/blob/ed52f0f01bed107c15c27d03f0bb4c89a010c1da/src/modules/auth/index.js#L88-L94

src/containers/Application.js
```diff
@@ -225,6 +232,46 @@ it is accessible by the sidebar toggler button.
...
+		// Add a request interceptor
+		axios.interceptors.request.use(
+			interceptor,
+			function(error) {
+				return Promise.reject(error)
+			});
+	}
+
+
+	interceptorResponse(axios, interceptor) {
```
ateska, line 245: This is unnecessary.
It is a normal error handling, not specific to interceptors.

401 is "Not authorized" ... Token expiration is just one specific reason when this could happen.

ateska (changes requested):

ateska (approved):

mpavelka (approved):

PR #85, opened 2021-03-19
INDG-94 - Isolate all places in the WebUI, where absolute URL is still needed - JSX, CSS, HTML, etc - and specify relative paths

titkov49:


ateska (approved):

Pe5h4 (approved):

titkov49:
@ateska @Pe5h4 just found out that webpack.dev had different output for js and css  (static/js and static/css respectively) so should had been changes as well
https://github.com/TeskaLabs/asab-webui/pull/85/commits/c373438d388c7cf7141c2498a608f7deb569b2c3

titkov49:
So can I merge it?

ateska:
Yes, let's merge it.

PR #86, opened 2021-03-22
Implement Access detail item into dropdown

Pe5h4:
Implementation of Access detail item into the auth header dropdown.

### DIFF
--- a/src/modules/auth/index.js
+++ b/src/modules/auth/index.js
@@ -4,6 +4,7 @@ import HeaderComponent from './header'
 import reducer from './reducer';
 import { types } from './actions'
 import { SeaCatAuthApi, GoogleOAuth2Api } from './api';
+import AccessControlScreen from './AccessControlScreen';
 
 
 export default class AuthModule extends Module {
@@ -22,6 +23,18 @@ export default class AuthModule extends Module {
 		this.App.addSplashScreenRequestor(this);
 
 		this.Authorization = this.Config.get("Authorization"); // Get Authorization settings from configuration
+
+		// Access control screen
+		app.Router.addRoute({
+			path: '/access',
+			exact: true,
+			name: 'Access control',
+			component: AccessControlScreen,
+			hasHeader: false,
+			hasSidebar: false,
+			hasBreadcrumb: false,
+			hasFooter: true,
+		});
 	}
 
 
--- a/src/modules/auth/index.js
+++ b/src/modules/auth/index.js
@@ -26,7 +26,7 @@ export default class AuthModule extends Module {
 
 		// Access control screen
 		app.Router.addRoute({
-			path: '/access-control',
+			path: '/auth/access-control',
 			exact: true,
 			name: 'Access control',
 			component: AccessControlScreen
--- a/src/modules/auth/index.js
+++ b/src/modules/auth/index.js
@@ -26,14 +26,10 @@ export default class AuthModule extends Module {
 
 		// Access control screen
 		app.Router.addRoute({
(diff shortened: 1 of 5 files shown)


ateska (approved):

ateska:
It is not 'translatable' right? ;-)

"Access control" is better name.

Pe5h4:
@ateska Well, everything is translatable, so I did it the way, that translations for this dropdown and for Access control screen can be added to the locales of the product (e.g. LogMan.io WebUI, SeaCat WebUI,...) and served from there. If it will not be provided in the locales of the product, the fallback will be the original non translated text. You can see the example of the implementation in the demo app of ASAB WebUI https://github.com/TeskaLabs/asab-webui/blob/06b1a7415774fb0212e579f2838b3f7578e13742/demo/public/locales/cs/translation.json and e.g. the auth header dropdown here: https://github.com/TeskaLabs/asab-webui/blob/06b1a7415774fb0212e579f2838b3f7578e13742/src/modules/auth/header.js#L35

However, I moved the Access control screen from SeaCat Auth WebUI to ASAB WebUI as you suggested. It is a part of `auth` module. Here you can see the working example with the translations on LM.io WebUI:

asab-webui-accesscontrol-screen

src/modules/auth/index.js
```diff
@@ -22,6 +23,18 @@ export default class AuthModule extends Module {
 		this.App.addSplashScreenRequestor(this);
 
 		this.Authorization = this.Config.get("Authorization"); // Get Authorization settings from configuration
+
+		// Access control screen
+		app.Router.addRoute({
+			path: '/access',
```
ateska, line 29: `/auth/access-control`

src/modules/auth/index.js
```diff
@@ -22,6 +23,18 @@ export default class AuthModule extends Module {
...
+			exact: true,
+			name: 'Access control',
+			component: AccessControlScreen,
+			hasHeader: false,
+			hasSidebar: false,
+			hasBreadcrumb: false,
+			hasFooter: true,
```
ateska, line 36: Do we still use these `has...` ?

ateska (changes requested):

src/modules/auth/index.js
Pe5h4: Oh, we dont

Pe5h4:
Path of the Access control screen updated. Removed obsolete parameters from navigation of Access control screen.

src/modules/auth/index.js
```diff
@@ -22,6 +23,14 @@ export default class AuthModule extends Module {
 		this.App.addSplashScreenRequestor(this);
 
 		this.Authorization = this.Config.get("Authorization"); // Get Authorization settings from configuration
+
+		// Access control screen
+		app.Router.addRoute({
+			path: '/access-control',
```
ateska, line 29: I would really like to have a path prefix here to avoid conflicts with the rest of the application space.

`/auth/access-control`

ateska (changes requested):

src/modules/auth/index.js
Pe5h4: Implemented

ateska (approved):

ateska (approved):

Pe5h4:
@ateska just fyi: I also added comment for devs about how to use language locales for auth Header and AccessControlScreen components https://github.com/TeskaLabs/asab-webui/pull/86/commits/fa1c11ce9e1bdc936bf0ec9214f9162c9cd9b1ca

PR #94, opened 2021-04-20
Refactor userinfo call to obtain roles and resources for active tenant

Pe5h4:
This PR refactor userinfo API call to obtain Roles and Resources from database for active tenant.


src/modules/auth/index.js
```diff
@@ -173,8 +173,12 @@ export default class AuthModule extends Module {
 
 	async _updateUserInfo() {
 		let response;
+		// Obtain tenant for API call from URL params
+		const search = window.location.search;
+		const params = new URLSearchParams(search);
+		let activeTenant = params.get('tenant');
```
ateska, line 179: `tenant` is optional ... if not present then it is not included in the userinfo call.

ateska (changes requested):
tenant in userinfo / application query string is of course optional

src/modules/auth/index.js
Pe5h4: @ateska Here it should not have any impact. If tenant is not in the url parameters, SeaCat Auth service is able to handle it (what I spoke of with Robin and from testing). This can mean that tenant in userinfo is not present or it can be also an empty list.

The API call looks like this

https://github.com/TeskaLabs/asab-webui/blob/8e4c8b8a05893897707b20adc761701572918ec6/src/modules/auth/api.js#L60

The `active_tenant` is send as a parameter in the request of the API call (not in the payload) and so when the `active_tenant` is undefined, null or empty string, the response is always without roles/resources/tenant in the userinfo.

But I might not see some things which will arise in the future. So I can separate the API call into 2 (one with `tenant` parameter in the API request and the other without it) and will use them accordingly, when active tenant from url params will be undefined.

Anyway, I did a bit of refactoring on tenant module to ensure that the application is loaded even when tenant is not in the userinfo response: https://github.com/TeskaLabs/asab-webui/pull/94/commits/8e4c8b8a05893897707b20adc761701572918ec6

ateska: Well, what I mean mainly is that `userinfo(access_token, active_tenant)` signature is wrong since it doesn't make sense in deployments that don't use tenant.

Just move the code handling tenant inside `userinfo()` method, then is it encapsulated (invisible to an user programmer).

ateska (changes requested):

src/modules/auth/index.js
Pe5h4: Oki, moved and tested https://github.com/TeskaLabs/asab-webui/pull/94/commits/2e331b4772d803f9affdaef0e1af6df2681fc4a9

src/modules/auth/api.js
```diff
@@ -53,11 +53,17 @@ export class SeaCatAuthApi {
...
+		const search = window.location.search;
+		const params = new URLSearchParams(search);
+		let active_tenant = params.get('tenant');
+		// Add access bearer token to the Authorization headers
 		if (access_token != null) {
 			headers.Authorization = 'Bearer ' + access_token;
 		}
-		return this.OidcAPI.get('/userinfo', {headers: headers});
+		// The userinfo API call will handle the cases where tenant is not available or undefined
+		return this.OidcAPI.get('/userinfo?tenant=' + active_tenant, {headers: headers});
```
ateska, line 66: Still - I don't want to generare request with "?tenant=" for deployments where tenants are not used.

ateska (changes requested):

src/modules/auth/api.js
Pe5h4: Oh, sure, sorry, I misunderstood. https://github.com/TeskaLabs/asab-webui/pull/94/commits/bfc6e1a7617e5a1b848137b132dc29db0d50d0a1

I also changed the URL of Access control screen (it is the screen which is displayed in the header dropdown with the information about accesses) when the tenant is not present. https://github.com/TeskaLabs/asab-webui/pull/94/commits/20fcbf75bc4fa935a963d6a021489bb2f218a31d

ateska (approved):

PR #97, opened 2021-04-24
Add a posibility to specify tenants in the static (build) configuration.

ateska:
I encountered a situation when I need tenants but there is not Auth.
For this - rather exotic - situation, I need to introduce tenants thru configuration.

### DIFF
--- a/src/modules/tenant/service.js
+++ b/src/modules/tenant/service.js
@@ -7,6 +7,20 @@ export default class TenantService extends Service {
 		super(app, name);
 	}
 
+	async initialize() {
+
+		// If the tenant list is in the configuration, use it
+		// This is useful when tenants are present but auth not
+		var tenants = this.App.Config.get('tenants');
+		if (tenants != null) {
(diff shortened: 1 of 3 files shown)


src/modules/tenant/service.js
```diff
@@ -7,6 +7,20 @@ export default class TenantService extends Service {
 		super(app, name);
 	}
 
+	async initialize() {
+
+		// If the tenant list is in the configuration, use it
```
Pe5h4, line 12: We would need to extend the documentation about tenant usage without authorization
https://github.com/TeskaLabs/asab-webui/blob/implement/add-tenants-config/doc/tenant.md

Pe5h4 (changes requested):

src/modules/tenant/service.js
ateska: Documentation extended.

Pe5h4 (approved):

PR #98, opened 2021-04-26
Remove moment/locales (except czech and english)

titkov49:
Moment.js package includes locales for itself. By removing most of it (I kept only czech and english) we can reduce size of bundle by 354KB (1020KB -> 666KB). Also I attach pictures of webpack-analyzer-plugin to show how big was those locales in compare with other packages
asab-webui-without-locales
asab-webui-with-locales

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -30,6 +30,7 @@
 	"style-loader": "^1.1.3",
 	"uglifyjs-webpack-plugin": "^2.0.1",
 	"webpack": "^4.42.0",
+	"webpack-bundle-analyzer": "^4.4.1",
 	"webpack-cli": "^3.3.11",
 	"webpack-dev-server": "^3.10.3",
 	"webpack-node-externals": "^1.7.2",
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
 		"style-loader": "^1.1.3",
 		"uglifyjs-webpack-plugin": "^2.0.1",
 		"webpack": "^4.42.0",
+		"webpack-bundle-analyzer": "^4.4.1",
 		"webpack-cli": "^3.3.11",
 		"webpack-dev-server": "^3.10.3",
 		"webpack-node-externals": "^1.7.2",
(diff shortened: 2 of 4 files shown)


titkov49:
Also added webpack-analyzer-plugin into package.json and commented it in webpack in case we will need to check asab-webui.

titkov49:
@ateska can you please check and say if we need some other locales for moment except cs and en

ateska:
I see many components that we basically don't need for a login form - please discuss that with the team.
Also, I guess, we need to find a way how to limit the inclusion of the libraries from ASAB WebUI modules into the compiled main.js of the final product (SeaCat Auth WebUI in this case).

ateska (changes requested):
Yes, we will need basically any possible locale - but that is configurable by the final build of the dependant application.
So basically we don't need to include any but there has to be a way how user/programmer can specify what locales he/she want to include.

wuujcik (approved):

titkov49:
> Yes, we will need basically any possible locale - but that is configurable by the final build of the dependant application.
> So basically we don't need to include any but there has to be a way how user/programmer can specify what locales he/she want to include.

@ateska Okay, now we can define moment locales in config file https://github.com/TeskaLabs/asab-webui/pull/98/commits/39a9590d3077e144ee035ef4b70352e155b9af62

demo/conf/config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
 	app: {
 		API_PATH: 'api',
 		modules: ["HomeModule"],
+		momentLocales: ["cs", "en-gb"]
```
ateska, line 5: can we have just "locales" ... that could be then used on more places.

ateska (changes requested):
Configurable locales are good idea - but that's not moments specific - it is broader and has to be aligned with i18n module.
Consult with @Pe5h4 - how current locales are built and how to include these in the config.

titkov49:
@ateska So since we decide to replace moment with date-fns and work on locales later, can I merge this PR?

ateska (approved):

ateska:
Yes

PR #101, opened 2021-04-29
Quick Fix - Add check for config[app][momentLocales]

titkov49:
App is crashing when there is no config[app][momentLocales]

This is temporarily solution. Will be changed once we replace moment with date-fns

### DIFF
--- a/webpack/webpack.dev.js
+++ b/webpack/webpack.dev.js
@@ -15,7 +15,7 @@ module.exports = {
 		const entry_path = path.resolve(config["dirs"]["src"], 'index.js');
 		const html_template_path = path.resolve(config["dirs"]["public"], 'index.html');
 		// TODO: This is temporary solution. It will be replaced by date-fns.
-		let momentLocales = /en-gb|cs/;
+		let momentLocales = /en-gb|cs/; // Default locales
 		if (config["app"]["momentLocales"]) {
 			momentLocales = new RegExp(Object.values(config["app"]["momentLocales"]).join("|"));
 		}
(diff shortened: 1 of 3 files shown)


webpack/webpack.dev.js
```diff
@@ -15,7 +15,10 @@ module.exports = {
 		console.log(config);
```
Pe5h4, line 15: Remove console.log()

Pe5h4 (changes requested):

ateska (changes requested):
Please add a comment about "this is temporary solution, will be replaced by XXX"

Pe5h4 (approved):

ateska (approved):

titkov49:
@Pe5h4 @ateska 
Sorry for bothering again. I found out that this line (https://github.com/TeskaLabs/asab-webui/pull/101/commits/264cf02c2b52337e23973f78e66661ce02c28a23#diff-23950346c0c899636c507c0d59db096cace7a938aca2950e241b0bd090ae334aR20) doesn't return ["en-gb", "cs"] if config["app"]["momentLocales"] is undefined and momentLocales variable remains undefined. It doesn't throw any error, it's just that contextReplacementPlugin doesn't remove any locales if second argument is undefined. So I added default locales other (more simple and more readable) way.  
Can I ask you for one more review please? https://github.com/TeskaLabs/asab-webui/pull/101/commits/51b93af35176e17d44ca573749fc898b94ad9463

ateska (approved):

Pe5h4 (approved):

PR #103, opened 2021-05-03
Refactor format obtaining from `$defs`  of the schema

Pe5h4:
This PR refactor obtaining the format of the input field from the `$defs` property of the schema. This PR can be merged only after the changes are also approved in the asab-config service.

### DIFF
--- a/src/modules/asab-config/ConfigEditor.js
+++ b/src/modules/asab-config/ConfigEditor.js
@@ -250,7 +250,6 @@ function ConfigSection(props) {
 				<CardBody>
 					{Object.keys(props.section.properties).map((item_name, idx) =>
 						// Decide what type of config item to render based on format
-						// TODO: `format` names taken from `$defs` will be handled like below after the approval on asab-config service
 						{switch(props.section.properties[item_name]['$defs'] ? Object.keys(props.section.properties[item_name]['$defs'])[0] : undefined){
 							case 'text': return(<ConfigItem
 													key={idx}
--- a/src/modules/asab-config/ConfigEditor.js
+++ b/src/modules/asab-config/ConfigEditor.js
@@ -10,16 +10,12 @@ import {
 } from "reactstrap";
 
 import {
-	ConfigItem,
 	NumberConfigItem,
-	UrlConfigItem,
-	EmailConfigItem,
-	PasswordConfigItem,
 	CheckBoxConfigItem,
 	RadioButtonConfigItem,
 	SelectConfigItem,
-	TextAreaConfigItem,
 	ConfigAdHocItem,
+	StringItems
 } from './ConfigFormatItems';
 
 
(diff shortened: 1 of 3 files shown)


Pe5h4:
@ateska This PR refactor obtaining the input field type from the `asab-config` service from `$defs` property of the schema.

schema_type_validation_defs

Selection_235

ateska:
Looks good.
The main remark here is that `$def` should not not be required ... use `format` since that is always present and `$def` contains only "extra" configuration required by UI.

src/modules/asab-config/ConfigEditor.js
```diff
@@ -244,8 +250,7 @@ function ConfigSection(props) {
 				<CardBody>
 					{Object.keys(props.section.properties).map((item_name, idx) =>
 						// Decide what type of config item to render based on format
-						// TODO: `format` names should be consulted with BE team
-						{switch(props.section.properties[item_name]['format']){
+						{switch(props.section.properties[item_name]['$defs'] ? Object.keys(props.section.properties[item_name]['$defs'])[0] : undefined){
```
ateska, line 253: The main remark here is that $def should not not be required ... use format since that is always present and $def contains only "extra" configuration required by UI.

ateska (changes requested):

src/modules/asab-config/ConfigEditor.js
Pe5h4: @ateska `$defs` is required only for generating input formats for UI (it is used for softcheck, if the values are passed to zookeeper in correct form), if it is not defined, the `text` input is used as default as you can see here:

https://github.com/TeskaLabs/asab-webui/blob/refactor/obtaining-input-formats-from-schemas/src/modules/asab-config/ConfigEditor.js#L317-#L323

So if no `$defs` is passed in the JSON schema, the text input field is considered as a default.

As we discussed and if I remember correctly, the input field `format` for UI should be defined in `$defs` because implementing format out of `$defs` would break rules of JSON schema. Maybe instead of `format` you meant `type` ? Because this is present in every  JSON schema.

Here is the example of JSON schema from Mithun. For UI there is no other way to obtain the information what input field to render than `$defs` (or `type`)

```
                "precise_error_handling": {
                    "$id": "#/properties/connection%3AESConnection/properties/precise_error_handling",
                    "type": "boolean",
                    "title": "The precise_error_handling schema",
                    "description": "If set to True, checkbox is checked. Otherwise false.",
                    "default": false,
                    "examples": [
                        true
                    ],
                    "$defs": {
                        "checkbox": {
                          "type": "boolean",
                          "format": "checkbox"
                        }
                    }
                }
```

Pe5h4: If you meant `type`, I will refactor accordingly and will use `$defs` as an additional info for UI

ateska: Well, I mean that if the "type" is boolean, I don't need to specify `$defs` in order to use checkbox.

The goal - as usual - is to keep these files as short as possible in majority of cases.

Pe5h4: Got it! Thank you

Pe5h4:
@ateska Updated input fields by the `type` and `$defs` (if defined in schema) are used as an additional help to recognize the input field type for UI. Default input format field for `string` type is `text`, for type `number` or `integer` the input format field is always `number`. For `boolean` type, the input field format is always `checkbox`. I have not used the input format field `radiobutton` even though it is already implemented. I would need to test it on some particular case on which I cannot thought of any now.

Pe5h4:
@ateska

ateska (approved):

PR #106, opened 2021-05-07
Change microservices endpoint from '/svcs' to '/microservices'

titkov49:


ateska (approved):

Pe5h4 (changes requested):
Put this change into changelog please, just in case (into the `Refactoring` section)

PR #108, opened 2021-05-14
Create application build descriptor

titkov49:


script/lib/createBuildDescriptor.js
```diff
@@ -0,0 +1,19 @@
...
+    let yarnLockContent = fs.readFileSync("yarn.lock");
+
+    let dir = "dist/";
+
+    if (config["dirs"] && config["dirs"]["dist"]) {
+        dir = config["dirs"]["dist"];
+        if (!dir.endsWith("/")) dir += "/";
+    }
+
+    fs.writeFile(dir + "app_descriptor.lock", yarnLockContent, (err) => {
```
ateska, line 13: Here I suggest to keep the name `yarn.lock` since that's the content.
Then you add the file with `app_version.xxx` with `git describe` output.

ateska (changes requested):

ateska (approved):

script/lib/createBuildDescriptor.js
```diff
@@ -0,0 +1,35 @@
...
+
+    fs.writeFile(dir + "yarn.lock", yarnLockContent, err => {
+        if (err) throw err;
+        console.log("yarn.lock descriptor has been created");
+    })
+
+    exec("git describe --abbrev=7 --tags --dirty=+dirty --always", (err, stdout, stderr) => {
+        if (err) {
+            console.error(err);
+            fs.writeFile(dir + "version.txt", stderr, err => {
```
Pe5h4, line 22: @titkov49 Just a question - I thought the file with version should have been a `json` file or `md`, so why `txt` ?

Pe5h4 (approved):

script/lib/createBuildDescriptor.js
ateska: It is a single line of text, you can read that from JavaScript ... so it is fine as a .txt

BTW- we should also create "About" page in the asab-webui that should display the version info to the user (from user drop down).

Pe5h4: Noted and ticket created

PR #109, opened 2021-05-17
Return DateTime multiplier for timestamp in seconds (by 1000) and add support of timestamp in milliseconds

titkov49:
Because of misunderstood I deleted multiplier by 1000 in DateTime component. On SCRUM meeting we decided that it should stay.

Screenshot 2021-05-17 at 14 49 52
Screenshot 2021-05-17 at 14 50 05

### DIFF
--- a/src/containers/DateTime.js
+++ b/src/containers/DateTime.js
@@ -17,7 +17,7 @@ The input `value` is in UTC and one of:
 
 * Data() object (Javascript)
 * "2020-08-22T00:00:00+00:00" String with ISO format of the day
-* Number with Unix timestamp
+* Number with Unix timestamp (in seconds)
 
 The date&time will be converted to a local timezone of the browser.
 
@@ -35,7 +35,12 @@ export function DateTime(props) {
 		)
 	}
 
-	const m = moment(props.value);
+	let m = null;
+	if (isNaN(props.value)) {
+		m = moment(props.value);
+	} else {
+		m = moment(props.value * 1000.0);
+	}
 
 	return (
 		<span className="datetime" title={m.fromNow()}>
--- a/src/containers/DateTime.js
+++ b/src/containers/DateTime.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React from 'react'
 import moment from "moment";
 
 /*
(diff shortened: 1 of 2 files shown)


ateska:
We discussed the "smart" detection of the unit based on the range of the value provided.

titkov49:
@ateska @Pe5h4 @wuujcik 
Now DateTime component define which type of timestamp (in seconds or in milliseconds) was passed base on length of this number. It of course has own limits. For timestamp in seconds it is max value which is **9999999999** (Nov 20, 2286 6:46 PM). For timestamp in milliseconds it is min value which is ** 10000000000** (Apr 26, 1970 6:46 PM).

titkov49:
Screenshot 2021-05-18 at 12 39 33

src/containers/DateTime.js
```diff
@@ -17,7 +17,8 @@ The input `value` is in UTC and one of:
 
 * Data() object (Javascript)
 * "2020-08-22T00:00:00+00:00" String with ISO format of the day
-* Number with Unix timestamp
+* Number with Unix timestamp in seconds (max value is 9999999999)
+* Number with timestamp in milliseconds (min value is 10000000000)
```
ateska, line 21: Add the example with `Date.now()` please - it is quite relevant here.

src/containers/DateTime.js
```diff
@@ -27,20 +28,21 @@ The default format is `lll` -> `Aug 22, 2020 1:13 PM`
...
-	if ((props.value === null) || (props.value === undefined)) {
+	if ((value === null) || (value === undefined)) {
 		return (
 			<span className="datetime">{' '}</span>
 		)
 	}
 
-	const m = moment(props.value);
+	const m = isNaN(value) ? moment(value) :
+		value.toString().length > 10 ? moment(value) : moment(value * 1000);
```
ateska, line 40: `value >= 9999999999` ??

ateska (changes requested):

ateska (changes requested):
One remark, then mergeable.

ateska (approved):

Pe5h4 (approved):

PR #117, opened 2021-06-07
Implement resource access validation

Pe5h4:
- This PR implement secondary validation by resource access on navigation


Pe5h4:
@ateska This PR implement softcheck on application's resource access. When the resource is specified as a property of the Navigation in the application's module, the users without the particular access will not see the module in the navigation Sidebar.

`removeItem` method has been implemented to Application.js file of ASAB WebUI to remove the item from navigation.

Here is the flow for users with `seacat:access` and the user without the `seacat:access` resource. The one who does not have the resource, he is no able to display the SeaCat WebUI's `Auth` module in the sidebar navigation.

access-on-resource

ateska (approved):

PR #119, opened 2021-06-09
Log out after expired session

titkov49:


titkov49:
Since we haven't figured out how to implement translations outside of react component, I think that it should be done with sidebar translation. I will create ticket for that for next sprint.

src/modules/auth/index.js
```diff
@@ -170,6 +173,27 @@ export default class AuthModule extends Module {
...
+
+		if (!that.UserInfo?.exp) {
+			setTimeout(that.logoutOnExpiredSession, 1000, that);
+		}
+
+		if (difference <= 0) {
+			that.logout();
+			return;
+		} else if (difference < 60000 && !alert) {
+			that.App.addAlert("warning", `Your session will expire in ${Math.floor(difference/1000)} seconds.`);
```
ateska, line 188: This has to be localisable.

src/modules/auth/index.js
```diff
@@ -170,6 +173,27 @@ export default class AuthModule extends Module {
...
+	logoutOnExpiredSession (that, al=false) {
+		const difference = moment(that.UserInfo.exp).valueOf() - moment.now();
+		let alert = al;
+
+		if (!that.UserInfo?.exp) {
+			setTimeout(that.logoutOnExpiredSession, 1000, that);
+		}
+
+		if (difference <= 0) {
+			that.logout();
```
ateska, line 185: Nope, don't do `logout`, it will be problematic for users.

src/modules/auth/index.js
```diff
@@ -170,6 +173,27 @@ export default class AuthModule extends Module {
 		});
 	}
 
+	logoutOnExpiredSession (that, al=false) {
```
ateska, line 176: It is not the logout after expired session - it is a warning that that the session is going to expire, or is expired.

ateska (changes requested):

src/modules/auth/index.js
titkov49: We have an issue with localisation outside of functional react components right now so I suppose it should be done with localisation of Sidebar components

src/modules/auth/index.js
titkov49: Okay, then I should change this to second notification that session has expired?

ateska: yes

titkov49:
**Changes:**
@ateska 
1. Function was renamed `logoutOnExpiredSession` -> `notifyOnExpiredSession `
2. `notifyOnExpiredSession` now is called inside `initialize` method instead of `updateUserInfo`
3. Function was refactored into calling two `setTimeout` functions instead of using recursion
4. Function is showing second notification instead of logging out

src/modules/auth/index.js
```diff
@@ -170,6 +172,22 @@ export default class AuthModule extends Module {
 		});
 	}
 
+	notifyOnExpiredSession () {
+		const difference = moment(this.UserInfo.exp).valueOf() - moment.now();
+		const that = this;
+
```
ateska, line 178: I don't think this is correct.

The session can be prolonged by the user activity - which means that this method needs to be called periodically and do the check on the userinfo everytime it "feels" that the session is going to expire or is expired.

Only if that's correct, then the warning is displayed ... and it has to be done in a way that the user is not getting multiple alerts at one; which means that there needs to be some kind of expiration of the alert set logically to align with a next step.

ateska (changes requested):

ateska:
Please also provide the gif with the functionality.

src/modules/auth/index.js
titkov49: As Robin told me there is no way to prolong user activity so I thought that this is alright.

ateska: Nope, it is not ... since the prolonging will be (obviously) added to SeaCat Auth soon.

ateska: Also, I would like to avoid moments since we want to depart from them, correct?

src/modules/auth/index.js
```diff
@@ -170,6 +172,26 @@ export default class AuthModule extends Module {
 		});
 	}
 
+	notifyOnExpiredSession (that, al=false) {
+		const userInfo = that.UserInfo;
+		let alert = al;
+
+		that._updateUserInfo();
```
ateska, line 179: This is the asynchronous method ... is this working?

src/modules/auth/index.js
ateska: And you cannot localize the message here?

src/modules/auth/index.js
```diff
@@ -97,6 +98,7 @@ export default class AuthModule extends Module {
 			}
 		}
 		this.App.removeSplashScreenRequestor(this);
+		this.notifyOnExpiredSession(this);
```
ateska, line 101: This will call update user info again at the start. That's incorrect b/c userinfo will be called anyway.

ateska (changes requested):
I still believe that the flow of this is incorrect.

The proper way:

- Timer to be set in userinfo call
- in the timer handler, you should check how old the user info is. If it is older than 30 seconds, do the refresh of user info (which will set the timer again)
- with the user info that is "fresher" than 30 seconds, you can do a check on the session expiration:
- if the session is expiring in next 60 seconds, show the alert **for exactly 60 seconds** - there is a option for that. Also, the alert should be displayed only once (this flag to be refreshed once the session lifetime is prolonged - eventually)
- if the session is expired, show the alert saying "session expired" and that's the end of this procedure.

titkov49:
@ateska @Pe5h4 re-review please :)

ateska (approved):

src/modules/auth/index.js
```diff
@@ -191,31 +190,58 @@ export default class AuthModule extends Module {
...
+		let tenants_list = that.UserInfo.tenants;
+		if (that.App.Services.TenantService) {
+			that.App.Services.TenantService.set_tenants(tenants_list);
+		}
+
+		/* Notifying user about session expiration section */
+		let exp = new Date(that.UserInfo.exp).getTime(); // Expiration timestamp
+		const difference = exp - Date.now(); // Difference between current time and expiration date in milliseconds
+
+		// If userinfo expiration date has been changed then first alert flag
```
Pe5h4, line 227: @titkov49 Can you wrap multiple line comments into `/* Your comment */` and do not comment every each line with `// Your comment` ? Just for sake of some structural styling

Pe5h4 (approved):
@titkov49 Note the comment I left about the style of the notes in the code

PR #123, opened 2021-06-15
Add i18n localisation outside of function components

titkov49:
Screenshot 2021-06-23 at 17 06 07
Screenshot 2021-06-23 at 17 07 35


titkov49:
@ateska btw, how should we translate word **Session**?

ateska:
> @ateska btw, how should we translate word **Session**?

`Relace`

ateska (approved):

PR #126, opened 2021-06-21
Add backend on subpath

Pe5h4:
- Fix path to locales when application is built on subpath
- Fix URL of Access control screen and of SeaCat Auth WebUI when app is on subpath

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,3 +70,7 @@
 - Assure, that BASE_URL is always an absolute URL. Assure, that API_PATH can be set as an absolute URL. (INDIGO Sprint 210514, [!111](https://github.com/TeskaLabs/asab-webui/pull/111))
 
 - Fix public path issue when apps are build on subpaths (INDIGO Sprint 210528, [!118](https://github.com/TeskaLabs/asab-webui/pull/118))
+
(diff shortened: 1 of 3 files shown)


Pe5h4:
@ateska This PR fix 2 `subpath` issues when the app is build on the subpath. 

- First is the language localisation issue when application is build on subpath. The change was to add a i18n `backend` with pathname, when the backend service is undefined and propagate the path of the locales.

- Second is the issue in `auth` Header dropdown, where the URL path for Access control screen and SeaCat Auth WebUI (Login page) has not been propagated correctly when the main application has been build on subpath
  - Access control screen. This has been fixed by adding a location pathname
   https://github.com/TeskaLabs/asab-webui/blob/4eb6742a8d024c6ca1843f2abacdebb7e344f007/src/modules/auth/header.js#L43-#L50
  - SeaCat Auth WebUI (Login page) has been fixed by obtaining the URL from the window location, if it has not been set in the `SERVICES` in the config file.
https://github.com/TeskaLabs/asab-webui/blob/4eb6742a8d024c6ca1843f2abacdebb7e344f007/src/modules/auth/header.js#L36-#L40
 
The functionality of the components, when app is not build on the subpath and/or the language locales are loaded from external folder, has not been affected.

Here is the config example used for build on subpath:
Selection_239

Here is the flow of the app, when it is build on the subpath and switching between languages and accessing Access control screen and SeaCat Auth WebUI:
fix-locales-and-urls-on-subpath

ateska (approved):

PR #127, opened 2021-06-22
Feature/wizard

titkov49:
Continuing to work on asab-webui wizard. Created new module for Wizard


ateska (approved):

src/abc/Module.js
```diff
@@ -1,6 +1,10 @@
 export default class Module {
     constructor(app, name) {
     	this.App = app;
+        this.Name = name;
+        this.Config = app?.config;
```
Pe5h4, line 5: @titkov49 Why do you need to define Config, Navigation and Router in here? Also, I think it should be without `?`. Also, be patient with the indentation on line 6 and 7

src/modules/wizard/WizardContainer.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { Wizard } from './Wizard';
+import { Container } from 'reactstrap';
+
+import json from './mock-data.json';
+
+const WizardContainer = (props) => {
+    console.log(json)
```
Pe5h4, line 9: @titkov49 Here is some forgotten console.log

src/modules/wizard/index.js
```diff
@@ -0,0 +1,23 @@
+import Module from 'asab-webui/abc/Module';
+import WizardContainer from './WizardContainer';
+
+export default class WizardModule extends Module {
+    constructor (app) {
+        super(app, "WizardModule");
+
+        console.log("WizardModule: ", this);
```
Pe5h4, line 8: @titkov49 Here is some forgotten console.log

src/modules/wizard/mock-data.json
```diff
@@ -0,0 +1,119 @@
+{
```
Pe5h4, line 1: @titkov49 Do we need mocked data in the `modules/wizard`? It might be better, if it would be a part of documentation or `public` folder

Pe5h4 (changes requested):

Pe5h4:
Closing for obsolescence

PR #136, opened 2021-07-08
Fix AccessControlScreen when userinfo is null (after expired session)

titkov49:
I suppose that this solution is better in case when session has expired rather than keep old userinfo b/c in that case user can't see any info after session has expired.
 
 Screenshot 2021-07-08 at 4 05 17

### DIFF
--- a/src/modules/auth/AccessControlScreen.js
+++ b/src/modules/auth/AccessControlScreen.js
@@ -48,7 +48,6 @@ function AccessControlCard(props) {
 	const { t, i18n } = useTranslation();
 	let history = useHistory();
 	let userinfo = props.userinfo;
-	
 	let App = props.app;
 	let currentTenant;
 	// Check Tenant service availability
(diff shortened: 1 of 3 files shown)


src/modules/auth/AccessControlScreen.js
```diff
@@ -65,32 +65,40 @@ function AccessControlCard(props) {
...
-				<Row>
-					<Col>
-						<h5>{t('AccessControlScreen|Resources')}</h5>
-					</Col>
-					<ItemToRender userinfo={userinfo} item='resources' />
-				</Row>
+					</>
+					) : (
+					<div className="text-center">
+						<h5>{t("AccessControlScreen|UserInfo doesn't exist.")}</h5>
```
ateska, line 99: The message is "weird", I suggest:
"The user information is invalid, you session is likely expired."

ateska (changes requested):

ateska (approved):

Pe5h4 (approved):

PR #139, opened 2021-07-19
Add condition on undefined release date

Pe5h4:
- Refactor About module with missing condition on undefined release date


Pe5h4:
@ateska In this PR, I refactored the `About` screen with condition when `release date` is undefined.

The previous state when `release date` is `undefined` was:
Selection_250

The current refactored state:
Selection_249

ateska (approved):
Yes, this is more defensive.
Note: The release date and the version should be filled by CI/CD.

PR #141, opened 2021-07-27
Refactor Sidebar

titkov49:


titkov49:
Screenshot 2021-07-30 at 15 58 55
Screenshot 2021-07-30 at 15 59 01

Pe5h4 (commented):
@titkov49 Cool! Is there any documentation about how to sort items in sidebar?

src/style.scss
```diff
@@ -26,3 +26,66 @@ main > .container-fluid, main > .container {
...
+			margin: 0;
+
+			.nav-link {
+				display: block;
+				.sidebar-item-icon {
+					display: inline-block;
+					width: 1rem;
+					margin: 0 0.5rem 0 0;
+					color: #73818f;
+					// font-size: 0.875rem;
```
Pe5h4, line 65: @titkov49 You forgot one comment here

titkov49: Yep, you're right. Sorry!

src/containers/Sidebar/SidebarItem.js
```diff
@@ -0,0 +1,63 @@
...
+										<Icon icon={child.icon} />
+										{t(`Sidebar|${child.name}`)}
+									</NavLink>
+								))}
+							</Nav>
+						</Collapse>
+					)
+				}
+			</NavItem>
+		</>
```
t-pyrope, line 59: I think React.Fragment is not necessary here

src/containers/Sidebar/index.js
```diff
@@ -0,0 +1,29 @@
...
+			<AppSidebar fixed display={props.display}>
+				<AppSidebarHeader />
+				<AppSidebarForm />
+				<SidebarNavItems
+					app={props.app}
+					navConfig={props.navigation.getItems().items}
+				/>
+				<AppSidebarFooter />
+			</AppSidebar>
+		</React.Fragment>
```
t-pyrope, line 25: I think `React.Fragment is not necessary here`

doc/sidebar.md
```diff
@@ -0,0 +1,16 @@
+# Changing order of sidebar navigation items
+
+It is possible to change order of sidebar navigation items with configuration of app. Just add items names to property `sidebarItemsOrder` as list in that order which you want it to appear in sidebar: 
+
+```
+let ConfigDefaults = {
+	...
+	sidebarItemsOrder: ["Home", "Configuration", "Dashboard"]
```
Pe5h4, line 8: @titkov49 is `sidebarItemsOrder` takes the name of the `items` in the `Module`, am I right?

Pe5h4 (approved):

doc/sidebar.md
titkov49: @Pe5h4 yep, that's right

ateska: Unfortunately, we cannot implement it this way - it assumes that application knows the final list of the sidebar items - which is non-systematic assumption.

It could be done e.g. by specifying a position "number" when adding item to the sidebar - then items are sorted by this number.

ateska (changes requested):

doc/sidebar.md
titkov49: I also was thinking about position number but if such would be implemented it would mean that in case if developer would need to change order of items in sidebar he might will need to go through each index.js of each module and change it's position which will make a lot of pain. Or if developer specify the same number for two modules than he can't control which of those two goes first and second which also isn't correct as I think.

titkov49: Okay, all that way along I was thinking about position number as integer but I understand now that position number can also be float which would solve issue that I described above. 
Anyway current ordering just throws all items that is not written in `sidebatItemsOrder` to the end of the list. In that way such ordering doesn't assume that it knows all items that appear or may appear in sidebar. It just sort them by list that is given from config. What I mean is that if we have three items in sidebar `["Home", "Table", "Dashboard"]` and in **sidebarItemsOrder** we specify only two from that list i.e. `sidebarItemsOrder=["Table", "Dashboard"]` then we will get sidebar with such ordering: `Table -> Dashboard -> Home`. 
If it isn't not what you meant then I can refactor that with position number.

t-pyrope (approved):

doc/sidebar.md
ateska: The position number is better for "composability" - e.g. in SeaCat WebUI, some modules are not loaded at all, so I don't know what is the final list of the items on the sidebar. But I can always define "approximate" order.
So if we say that you need to use number from 0-1000, then you know that small numbers are at the top and close to 1000 will be on the bottom. That will be fine for our goals.

titkov49:
@ateska @Pe5h4 @t-pyrope re-review with latest changes please:
https://github.com/TeskaLabs/asab-webui/pull/141/commits/df69225bc581bb0ac56b517538ff8e8fd1944979

doc/sidebar.md
```diff
@@ -0,0 +1,45 @@
+# Changing order of sidebar navigation items
+
+It is possible to change order of sidebar navigation items when you add items to Navigation. Just add `order` property with number value to object which you pass into `app.Navigation.addItem()` and sidebar items will be automatically sorted from smallest to highest: 
+
+```
+	app.Navigation.addItem({
+		name: 'Nowhere',
+		url: '/nowhere',
+		icon: 'cil-casino',
+		order: "3"
```
ateska, line 10: Use integers in the example ...

ateska (changes requested):
Small remark, then mergeable.

Pe5h4 (approved):

t-pyrope (approved):

PR #143, opened 2021-07-28
Fix microservices containers indentation 

titkov49:


Pe5h4 (approved):
@titkov49 Please fix it also in `microservice.css` file

PR #145, opened 2021-07-29
Extend DataTable header by option to add a custom button

titkov49:


src/containers/DataTable/index.js
```diff
@@ -70,14 +71,32 @@ export function DataTable ({
 							<i className={title.icon}></i> : title.icon
 						}
 						{title.text}
+						{button &&
```
Pe5h4, line 74: @titkov49 What if I would like to have a dropdown here or something else? Is it possible to create the component customizable?

Pe5h4 (changes requested):
@titkov49 Missing changelog

src/containers/DataTable/index.js
```diff
@@ -70,14 +71,32 @@ export function DataTable ({
...
 							<div className="float-right ml-3 data-table-create-button">
 								<Link to={{ pathname: createButton.pathname }}>
 									<Button tag="span" size="sm">
 										{createButton.icon && 
-										<span className="pr-1">
-											{typeof createButton.icon === 'string' ? <i className={createButton.icon}></i> : createButton.icon}
-										</span>
+											<span className="pr-1">
+												{typeof createButton.icon === 'string' ? <i className={createButton.icon}></i> : createButton.icon}
+											</span>
```
t-pyrope, line 99: I think indents are not correct here, because GitHub makes them green

titkov49: They were not correct because they were aligned with parent curly brackets, I changed them on purpose.

src/containers/DataTable/index.js
titkov49: I think It would be better to create custom dropdown instead

Pe5h4: Yes, I think we would use that. If the dropdown can be customized (e.g. if I can modify the props of dropdown items), it would be great

titkov49:
@Pe5h4 @t-pyrope re-review please :)

Pe5h4 (approved):

Pe5h4 (changes requested):
@titkov49 Still missing CHANGELOG

titkov49:
> @titkov49 Still missing CHANGELOG

@Pe5h4 Edited

Pe5h4 (approved):

ateska (approved):
Approved - but be careful with this "Custom..." elements - it is better to allow the user to include any React component.

PR #147, opened 2021-08-02
Add split functionality to custom dropdown

titkov49:


Pe5h4:
@titkov49 Dima, would not be better to allow devs to pass whatever react component to the header as we discussed? You can create a ticket for that and work on it next sprint

PR #148, opened 2021-08-03
Refactor config service

titkov49:
Config Service gets config url from meta with name `x-config`

### DIFF
--- a/src/config/ConfigService.js
+++ b/src/config/ConfigService.js
@@ -40,21 +40,13 @@ export default class ConfigService extends Service {
 
 
 	initialize() {
-		let config_url = this.App.Config.get('CONFIG_PATH');
-
-		const xConfig = document.getElementsByName('x-config')[0];
-		console.log(xConfig.content);
-
-		// Provide backward compatibility with CONFIG_URL
-		if (this.App.Config.get('CONFIG_URL')) {
-			config_url = this.App.Config.get('CONFIG_URL');
-		}
+		const X_Config = document.getElementsByName('x-config')[0].content;
 
 		// Check on undefined configuration
-		if (config_url !== undefined) {
+		if (X_Config !== undefined) {
 			this.App.addSplashScreenRequestor(this);
 			let axios = Axios.create({baseURL: window.location.protocol + '//' + window.location.host});
-			axios.get(config_url).then(response => {
+			axios.get(X_Config).then(response => {
 				// Check on status and content-type
 				if ((response.status === 200) && (response.headers["content-type"] !== undefined && response.headers["content-type"].includes("application/json"))) {
 					this.Config._dynamic_config = response.data;
--- a/src/config/ConfigService.js
+++ b/src/config/ConfigService.js
@@ -40,7 +40,7 @@ export default class ConfigService extends Service {
 
 
 	initialize() {
-		const dynamic_config_url = document.getElementsByName('x-config')[0].content;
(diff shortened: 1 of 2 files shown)


titkov49:
@ateska 
At this point I handled only extracting meta and it's content from header. I tried `import` and `require()` to get `./my-config.json` but faced some issues with that. And I just understood that content inside meta element should be url with which I should download configs with axios, am I right?

src/config/ConfigService.js
```diff
@@ -42,6 +42,9 @@ export default class ConfigService extends Service {
 	initialize() {
 		let config_url = this.App.Config.get('CONFIG_PATH');
 
+		const xConfig = document.getElementsByName('x-config')[0];
```
ateska, line 45: `X-Config`

ateska (changes requested):
Good!

ateska:
@titkov49 the goal is something like:

```
let config_url = document.getElementsByName('X-Config')[0];
```

The download of the config is already implemented.

src/config/ConfigService.js
titkov49: Dashes are not allowed in naming in javascript (parser assume that it is subtract operator). I can use underscore instead.

titkov49:
@ateska I removed old code for now. But maybe I should keep it in order to keep support of old configuration?

src/config/ConfigService.js
ateska: I meant:
```
const xConfig = document.getElementsByName('X-Config')[0];
```

src/config/ConfigService.js
```diff
@@ -40,18 +40,13 @@ export default class ConfigService extends Service {
 
 
 	initialize() {
-		let config_url = this.App.Config.get('CONFIG_PATH');
-
-		// Provide backward compatibility with CONFIG_URL
-		if (this.App.Config.get('CONFIG_URL')) {
-			config_url = this.App.Config.get('CONFIG_URL');
-		}
+		const X_Config = document.getElementsByName('x-config')[0].content;
```
ateska, line 43:
```
const dynamic_config_url = document.getElementsByName('x-config')[0].content;
```

Local variables are always in lowercase + underscores.

ateska (changes requested):
Styling remarks, then mergeable.

ateska:
> @ateska I removed old code for now. But maybe I should keep it in order to keep support of old configuration?

No.

Pe5h4 (changes requested):
@titkov49 Add a changelog and info and the info to the header of this PR.

src/config/ConfigService.js
```diff
@@ -40,18 +40,13 @@ export default class ConfigService extends Service {
 
 
 	initialize() {
-		let config_url = this.App.Config.get('CONFIG_PATH');
-
-		// Provide backward compatibility with CONFIG_URL
-		if (this.App.Config.get('CONFIG_URL')) {
-			config_url = this.App.Config.get('CONFIG_URL');
-		}
+		const dynamic_config_url = document.getElementsByName('x-config')[0]?.content;
```
ateska, line 43: Can you add as a comment the example of the tag used for specification of the documentation? @titkov49 

Also this needs to be documented somewhere ...

titkov49: @ateska 
Done: https://github.com/TeskaLabs/asab-webui/pull/149

PR #150, opened 2021-08-09
Remove children based on resource from sidebar

Pe5h4:
- Implement `removeChildren` method to remove children of items from sidebar based on resource.
- It is similar to removal of item from sidebar based on resource


Pe5h4:
@ateska This PR is about removal of children(s) from Sidebar.

It is a softcheck with usage of `rbac` endpoint and it should prevent to display a `forbidden`/`restricted access` children of item(s) in the sidebar menu to users without particular rights (resource).

Previously, the similar concept has been implemented for `items` in sidebar, this is just a extension that we can control what we want to display to users based on their rights.

In the `gif`

- the first user `NewUser` does not have rights to display whole `Auth` module of SeaCat WebUI
- the second user `Test3` has rights to display `Auth` module, but does not have right to display `Session` containers (a child of `Auth` item)
- the third use `Franta` has all the rights and can display `Auth` module and all its children, in this case, this user is a superuser.

remove-children

Pe5h4:
@ateska

src/containers/Application.js
```diff
@@ -570,6 +570,15 @@ class Navigation {
 	removeItem(item) {
 		this.Items = this.Items.filter(i => i.name != item.name);
 	}
+
+	removeChildren(children) {
```
ateska, line 574: I don't think that this is a correct way how to implement this.

Technically, you don't remove items/children but you hide them; or it could be implemented as hidden implicitly and shown when resource is granted.

ateska (changes requested):

Pe5h4:
@ateska So, I refactored it. The functionality is the same as in gif above, so I did not created a new one.
What I changed was:

- I deleted item and children removal from navigation in Application.js
- I updated `auth` module where unauthorized items and children of navigation are now stored in redux store
- Updated `Sidebar` that the nav items and nav children stays hidden (they are not displayed) if it match the criteria of `unauthorized`. The decision what is authorized/unauthorized is done by `rbac` endpoint.
- When there is no `auth` module enabled in the application, all the sidebar items and children are displayed

Pe5h4:
@titkov49 ^^^

titkov49 (approved):

Pe5h4:
@ateska Encapsulated the navigation validation for Sidebar into separate async method (based on the review). The bulk API request to `rbac` endpoint will be implemented in the next iteration (after discussion with Robin)

Pe5h4:
@ateska

src/modules/auth/actions.js
```diff
@@ -1,3 +1,5 @@
 export const types = {
 	AUTH_USERINFO: "SeaCat/AUTH_USERINFO",
+	UNAUTHORIZED_ITEM: "SeaCat/UNAUTHORIZED_ITEM",
```
ateska, line 3: This is way to generic ... `SeaCat/Navigation/UNAUTHORIZED_ITEM` is likely more accurate

src/modules/auth/index.js
```diff
@@ -188,6 +179,36 @@ export default class AuthModule extends Module {
 		});
 	}
 
+
+	async validateNavigation() {
+		let getItems = this.Navigation.getItems();
```
ateska, line 184: `this.Navigation` should be replaced by `this.app.Navigation` ... it respects the ownership.

src/modules/auth/reducer.js
```diff
@@ -2,6 +2,8 @@ import { types } from './actions'
 
 const initialState = {
 	userinfo: null,
+	unauthorizedItem: [],
```
ateska, line 5: `unauthorizedNavigationItem` ... and so on

ateska (changes requested):

src/modules/auth/index.js
Pe5h4: @ateska I suppose it does, its a global value set in the constructor: https://github.com/TeskaLabs/asab-webui/blob/fad6b15f8e26db2a4017f831dbcac5562cf0da59/src/modules/auth/index.js#L25

Anyway, if it should be changed by your proposal, then we have to refactor it elswhere, e.g.:
https://github.com/TeskaLabs/asab-webui/blob/fad6b15f8e26db2a4017f831dbcac5562cf0da59/src/modules/auth/index.js#L19-L21

or not?

ateska: Well, yes ... it is a better practice to use `this.App.XYZ` instead of `this.XYX = this.App.XYZ` because people knows the app structure from elsewhere.

Pe5h4:
@ateska The naming convention has been updated

ateska (approved):

PR #154, opened 2021-08-17
DataTable - Header custom component

titkov49:
`customDropdownButton` will be replaced with this `customComponent`

### DIFF
--- a/demo/src/modules/home/containers/TableContainer.js
+++ b/demo/src/modules/home/containers/TableContainer.js
@@ -249,32 +249,24 @@ export default function (props) {
 		}
 	}
 
-	const customDropdownButton = {
-		text: "Custom drop",
-		color: "success",
-		header: "header",
-		items: [
-			{
-				text: "First item",
-				props: {
-					onClick: () => alert("First item was clicked!"),
-					className: "first-item-classname"
-				}
-				
-			},
-			{
-				text: "Second item",
-				props: {
-					onClick: () => alert("Second item was clicked!"),
-					className: "second-item-classname",
-					style: {
-						backgroundColor: "black",
-						color: "white"
-					}
-				}
-			}
-		]
+	const nextPage = () => {
+		if (page*limit < count) {
+			setPage(prev => prev+1)
+		}
+		else {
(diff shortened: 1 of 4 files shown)


src/containers/DataTable/index.js
```diff
@@ -15,7 +15,7 @@ import {
 
 import Table from './Table';
 import Pagination from './Pagination';
-import CustomDropdownButton from './CustomDropdownButton';
+// import CustomDropdownButton from './CustomDropdownButton'; DON'T REMOVE YET. IT MAY BE USEFUL ON REFACTORING DATATABLE
```
t-pyrope, line 18: I would write `//` also before `DON'T REMOVE YET` for easy uncommenting
`// import CustomDropdownButton from './CustomDropdownButton'; // DON'T REMOVE YET. IT MAY BE USEFUL ON REFACTORING DATATABLE`

demo/src/modules/home/containers/TableContainer.js
```diff
@@ -249,33 +249,25 @@ export default function (props) {
...
-					}
-				}
-			}
-		]
+	const nextPage = () => {
+		if (page*limit < count) {
+			setPage(prev => prev+1)
+		}
+		else {
+			alert("You're on the last page");
```
t-pyrope, line 257: I think it would be better not to alert, but to disable button

ateska (approved):

t-pyrope (approved):

src/containers/DataTable/index.js
titkov49: That's okay. I will be working on this probably next sprint and will delete this myself.

demo/src/modules/home/containers/TableContainer.js
titkov49: That's just demo in order to show how developer can use this `customComponent`. I suppose that's not needed to write there some real styling logic

titkov49: I think this would make code more complex and if someone will take a look on it in order to understand how to use it, then he may be a bit confused with all that stuff.

PR #156, opened 2021-08-23
Add support to absolute links in logo

t-pyrope:


src/containers/Header.js
```diff
@@ -17,7 +17,46 @@ import HelpButton from '../helpButton';
...
+				full={{
+					src: props.brand_image.full,
+					alt: props.title,
+					width: 120,
+					height: 30,
+				}}
+				minimized={{
+					src: props.brand_image.minimized,
+					alt: props.title,
+					width: 30,
```
titkov49, line 54: In both returns there are almost the same things. Do not repeat yourself. 
In this case you may rewrite this without repeating like this:
Common props in `AppNavbarBrand` you may define in object which you will destructure in both returns.
```
const appNavbarBrandProps = {
  full: {
	  src: props.brand_image.full,
	  alt: props.title,
	  width: 120,
	  height: 30,
  },
  minimized: {
	  src: props.brand_image.minimized,
	  alt: props.title,
	  width: 30,
	  height: 30,
  }
}
...
if (...) <Link...><AppNavbarBrand {{ ...appNavbarBrandProps } ... /></Link>
return <AppNavbarBrand tag={'div'} { ...appNavbarBrandProps } />
```

titkov49 (changes requested):

titkov49 (approved):

PR #157, opened 2021-08-24
Refactor/Move pagination to separate component

t-pyrope:


titkov49 (changes requested):
You also need to export this component from src/index.js to make it reusable outside of asab-webui:
https://github.com/TeskaLabs/asab-webui/blob/master/src/index.js

titkov49 (approved):

PR #160, opened 2021-08-26
Sidebar minimize button

titkov49:
CHANGES:
- Return sidebar minimize (collapse) button
- Placed About module to the bottom of sidebar items list
- Fix bug with wrapping items when there are to many of them in sidebar (or when height of the screen is small)
- Change background of collapsed item when it is wrapped

**Note:**
I can't place About module's item just above minimize button because for some reason in that case collapse button disappear (some strange core-ui logic, it doesn't move anywhere, it literally disappears even if I wrap collapse button with simple div) 

https://user-images.githubusercontent.com/37152497/130993478-64c46721-4647-47a2-8fd7-35b938654ead.mov

### DIFF
--- a/src/containers/Sidebar/SidebarNavItems.js
+++ b/src/containers/Sidebar/SidebarNavItems.js
@@ -4,7 +4,9 @@ import { connect } from 'react-redux';
 import { Nav } from 'reactstrap';
 import SidebarItem from './SidebarItem';
 
-const SidebarNavItems = ({ navConfig, unauthorizedNavItems, unauthorizedNavChildren, sidebarItemsOrder }) => {
+const SidebarNavItems = ({
+	navConfig, unauthorizedNavItems, unauthorizedNavChildren
+}) => {
 
 	// Sort items based on config
 	const memoizedItemsList = useMemo(() => {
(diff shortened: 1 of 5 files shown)


src/containers/Sidebar/SidebarNavItems.js
```diff
@@ -4,7 +4,9 @@ import { connect } from 'react-redux';
 import { Nav } from 'reactstrap';
 import SidebarItem from './SidebarItem';
 
-const SidebarNavItems = ({ navConfig, sidebarItemsOrder }) => {
+import { AppSidebarMinimizer } from '@coreui/react';
```
Pe5h4, line 7: @titkov49 This component will let us reliant on `coreui` again and I suppose this is not intended. Can we build the minimizer component by ourselves?

Pe5h4 (changes requested):

src/containers/Sidebar/SidebarNavItems.js
titkov49: I suppose we can't because we use core-ui layout and Sidebar relies on body, Header and Footer. This would require to rewrite them as well in order to view everything correct. I don't think that it's really hard but still this is more complex task

Pe5h4 (approved):

ateska (approved):
Ok.

But I fully agree with comments ... we will likely depart from core-ui at one point, so this will need to be refactored (then).

PR #161, opened 2021-08-30
Add maxWidth to DataTable cell

t-pyrope:
maxWidth

### DIFF
--- a/src/containers/DataTable/Table.js
+++ b/src/containers/DataTable/Table.js
@@ -78,21 +78,26 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
 		cell = <>{icon} {cell}</>;
 	}
 
+	const style = {
+		whiteSpace: "nowrap",
+		maxWidth: "260px",
+		textOverflow: "ellipsis",
+		overflow: "hidden"
+	}
+
 	return idx === 0 ? (
 			<th
 				className="data-table-th"
 				scope="row"
-				style={{
-					whiteSpace: "nowrap",
-					maxWidth: "40rem",
-					textOverflow: "ellipsis",
-					overflow: "hidden" 
-				}}
+				style={ style }
 			>
 				{cell}
 			</th>
 		) : (
-			<td className="pl-3 data-table-td" style={{ whiteSpace: "nowrap" }}>
+			<td
+				className="pl-3 data-table-td"
+				style={ style }
+			>
 				{cell}
 			</td>
 		);
--- a/src/containers/DataTable/Table.js
+++ b/src/containers/DataTable/Table.js
@@ -18,11 +18,14 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
 		whiteSpace: "nowrap",
 		maxWidth: "40ch",
 		textOverflow: "ellipsis",
-		overflow: "hidden"
+		overflow: "hidden",
+		marginBottom: 0
 	}
 
 	if (header?.icon) {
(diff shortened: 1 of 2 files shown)


src/containers/DataTable/Table.js
```diff
@@ -78,21 +78,26 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
 		cell = <>{icon} {cell}</>;
 	}
 
+	const style = {
+		whiteSpace: "nowrap",
+		maxWidth: "260px",
```
titkov49, line 83: Change this value to `30ch` or `40ch`

src/containers/DataTable/Table.js
```diff
@@ -78,21 +78,26 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
...
 			<th
 				className="data-table-th"
 				scope="row"
-				style={{
-					whiteSpace: "nowrap",
-					maxWidth: "40rem",
-					textOverflow: "ellipsis",
-					overflow: "hidden" 
-				}}
+				style={ style }
```
titkov49, line 92: You shouldn't add maxWidth and ellipsis here. Add it only for links (line 50) and for common text (line 75).

src/containers/DataTable/Table.js
```diff
@@ -78,21 +78,26 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
...
 			>
 				{cell}
 			</th>
 		) : (
-			<td className="pl-3 data-table-td" style={{ whiteSpace: "nowrap" }}>
+			<td
+				className="pl-3 data-table-td"
+				style={ style }
```
titkov49, line 99: You shouldn't add maxWidth and ellipsis here. Add it only for links (line 50) and for common text (line 75).

titkov49 (changes requested):

src/containers/DataTable/Table.js
t-pyrope: Are you sure about line 75 ? It's `else cell = obj[header.key] ? obj[header.key] : "-";` on it

t-pyrope: Oh, I see now! I was not right, sorry

titkov49: @t-pyrope yes, I'm sure. You should wrap `obj[header.key]` and `"-"` into some span and add styles there. Otherwise if you add styles on line 92 you will apply this styles to all other components (customComponent and datetime as well, but we don't want them to be limited by width)

src/containers/DataTable/Table.js
```diff
@@ -47,7 +54,7 @@ const TableCell = ({ obj, header, idx, showJson, jsonTheme }) => {
 
 	else if (header.link) {
 		const pathname = header.link.pathname + obj[header.link.key];
-		cell = obj[header.key] ? <Link to={{ pathname }} className="data-table-link">{icon} {obj[header.key]}</Link> : "-";
+		cell = obj[header.key] ? <Link style={ textLinkStyle } to={{ pathname }} className="data-table-link">{icon} {obj[header.key]}</Link> : "-";
```
titkov49, line 57: May I also ask you to break lines between props. We should break lines when there are 3 or more props for one component. If we write oneliners we may make code unreadable. BTW I always forget about this myself.
Otherwise it's good and approved.

titkov49 (approved):

Pe5h4 (approved):

PR #162, opened 2021-08-31
Refactor svcs attention required in list

titkov49:
Refactor svcs attention required based on workshop with Mithun and Ales


ateska (approved):

titkov49:
> Seeing this again - attentions are a dictionary - key is a `att_id` and the value is a dictionary with at least `_c` (creation time).

@ateska 
Currently `attention_required` is an array of dictionaries (each dictionary has `id` which is `att_id` we discussed). Currently we don't have creation time. I will tell Mithun about that.

PR #166, opened 2021-09-06
Add option to hide the button instead of disabling it

Pe5h4:
- This PR updates `ButtonWithAuthz` component to allow hiding the button completely when the button is disabled


Pe5h4:
@ateska This PR is about adding an option to hide the `ButtonWithAuthz` component completely instead of disabling it. The default value for `hideOnUnauthorizedAccess` is `false` so when the `hideOnUnauthorizedAccess` feature is not set to `true`, then the button stays visible (and eventually disabled)

Here the user have the resource to `delete`, so the button is visible, but the buttons for `reset` and `suspend` stays hidden since the user does not have the rights to do so:
Selection_278

Pe5h4:
@ateska

ateska (approved):

PR #168, opened 2021-09-07
Extract tenant from URL params if not set in the redux store

Pe5h4:
This PR refactor obtaining current tenant. If current tenant not yet set in the redux store, it will be taken from the URL params.


Pe5h4:
@ateska

ateska (approved):

PR #169, opened 2021-09-10
Add AttentionCard component into MicroserviceDetailContainer

titkov49:
Instead of using `react-json-viewer` component `JSONTable` component was created in order to skip indexes in parent table and to render dates with `DateTime` component.
Screenshot 2021-09-10 at 15 31 52

Screenshot 2021-09-10 at 15 30 58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@
 
 - Add minimize button to Sidebar (INDIGO Sprint 210820, [!160](https://github.com/TeskaLabs/asab-webui/pull/160))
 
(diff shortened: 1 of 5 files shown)


titkov49:
@ateska @Pe5h4 review please

Pe5h4 (approved):
Looks good!

Pe5h4:
We might do a reusable component from that in the next iterations

ateska:
Yes, I agree ... this looks cool.

ateska (approved):

PR #171, opened 2021-09-14
AboutScreen — Hide row with empty version

t-pyrope:
Hide a row with the version in UserInterface card, when there is no version provided

hide-version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,6 +124,8 @@
 
 - Add a possibility to add icons to items of ActionButton component of DataTable (INDIGO Sprint 210903, [!173](https://github.com/TeskaLabs/asab-webui/pull/173))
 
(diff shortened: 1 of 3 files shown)


titkov49 (approved):

Pe5h4:
@t-pyrope Can you also hide website, when it will not be set?
https://github.com/TeskaLabs/asab-webui/blob/master/src/modules/about/AboutCard.js#L50

Pe5h4 (changes requested):
@t-pyrope Can you also hide website please, when it is not set?

https://github.com/TeskaLabs/asab-webui/blob/master/src/modules/about/AboutCard.js#L50

t-pyrope:
about

Pe5h4 (approved):

PR #176, opened 2021-09-17
Implement universal reusable Knowledge base component

titkov49:


ateska (approved):
As discussed on the review, this is OK ... but we will likely need to deal with absolute paths in the `index.json`.

Pe5h4 (changes requested):
Can you please add changes into CHANGELOG file? Also some printscreens would be cool

ateska:
A little bit of documentation (e.g. example of use) would be nice here.

titkov49:
@Pe5h4 re-review please :)

doc/knowledge-base.md
```diff
@@ -0,0 +1,194 @@
...
+```
+import React from 'react';
+
+import { KnowledgeBase } from 'asab-webui';
+
+const KnowledgeBaseContainer = (props) => {
+	return (
+		<KnowledgeBase
+			app={props.app}
+			api="content"
```
Pe5h4, line 121: @titkov49 `api` might be misleading here. I would recommend to use something like `apiPath` or something with informative value

titkov49: @Pe5h4 `api` changed to `apiPath`

Pe5h4 (approved):

PR #178, opened 2021-09-22
Refactor authorization

Pe5h4:
- All rbac calls has been removed from ASAB WebUI
- Authorization calls for rbac has been replaced by softcheck done on FE side (authorization of user by tenant, authorization items in sidebar) - so no rbac calls at all
- Authorization by tenant is now enabled by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,10 @@
 
 - Update enabling ButtonWithAuthz component when user has superuser rights (INDIGO Sprint 210917, [!177](https://github.com/TeskaLabs/asab-webui/pull/177))
 
+- Remove RBAC endpoints completely from ASAB WebUI Auth module (INDIGO Sprint 210917, [!178](https://github.com/TeskaLabs/asab-webui/pull/178))
+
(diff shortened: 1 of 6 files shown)


Pe5h4:
@ateska In this PR

- all `rbac` calls has been removed from ASAB WebUI
- authorization calls for `rbac` has been replaced by `softcheck` done on FE side (authorization of user by tenant, authorization items in sidebar) - so no `rbac` calls at all
- authorization by tenant is now enabled by default (if user/developer would like to disable it, it has to be done in config or config defaults of the app) and only for cases when Tenant module is enabled in the application
- after this PR will be approved, I will change the docu on WIKI http://wiki.teskalabs.int/rd/projects/asab-webui/configuration/authorization , change configuration in all of our products (I will just remove obsolete authorization setting) and inform devs in `dev-all` channel
- the "breaking" change in the config files is just removal of obsolete authorization by tenant setting from
```
app: {
	...
	authorization: {
		authorize: true,
		resource: 'tenant:access',
		unauthorized_logout_timeout: 60000
	},
	...
},
```
to
```
app: {
	...
},
```
However, this "breaking" change has no effect on applications with authorization configured the old way as described above - What it means is, that if we will not remove obsolete authorization configuration, it will not break the application and the authorization will stay enabled.

How to disable authorization is described here: https://github.com/TeskaLabs/asab-webui/blob/bf3f766b39cc0754140351c07bb5b2afbf0da899/doc/authorization.md

And I will also put it to WIKI.

authorization

ateska (approved):

PR #179, opened 2021-09-24
Increase expiration alert duration to 1000 hours

titkov49:
Increase expiration alert (alert about expired session) duration to 1000 hours


ateska (approved):

Pe5h4 (approved):
Just add to the description, that it is a expiration for session expiration alert. And put it to the changelog

PR #182, opened 2021-10-05
Implementation of Tenant selector in splashscreen

Pe5h4:


Pe5h4:
@ateska In this PR tenant selection card has been implemented for users which will not pass tenant authorization. User can select between their available tenants. IIf user does not have any tenant assigned the tenant card selector will not show up and user's access is not granted. To enable (display) the tenant selector card, application must fulfill several conditions:

- tenant service must be present
- auth module must be enabled
- authorization by tenant must be enabled
- tenants must be present in the userinfo
- user can't be `superuser` (superuser will pass through any tenant authorization, so tenant selector does not have a meaning for `superusers`)
- the current tenant must be invalid for the user

Here is the visual part:
tenant-selector-card

src/modules/tenant/service.js
```diff
@@ -63,6 +64,7 @@ export default class TenantService extends Service {
 			type: types.TENANTS_CHANGED,
 			tenants_list,
 			current: current_tenant,
+			invalid: invalid
```
ateska, line 67: No, that's unnecessary and incorrect state.
Don't emit "tenant changes" when the authorisation fails - from the code, it seems to me that you don't need to have it at all.

ateska (changes requested):

PR #183, opened 2021-10-11
Move about module below to the bottom of the sidebar

titkov49:


titkov49:
Changes: 
1) Edit changelog
2) Align text in sidebar item to left
3) Use `Array.prototype.filter` instead of `Array.prototype.find`

PR #186, opened 2021-10-18
Implement ErrorHandler

titkov49:


titkov49:
Added two ErrorHandlers:
1) One wraps our JSX (excluding Header) - this handles errors upper in the tree
2) Second wraps containers (that excludes Header, Sidebar and Footer) - this handles only errors in modules

ateska:
Looks cool.
Eventually we can integrate some kind of "Submit the error" feature.

PR #193, opened 2021-11-02
Add possibility to disable breadcrumbs feature in specific container

titkov49:
This feature adds possibility to disable breadcrumbs in specific container in case if developer want so.

### DIFF
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -37,7 +37,7 @@ import { ADD_ALERT, SET_ADVANCED_MODE, CHANGE_HELP_URL } from '../actions';
 class Application extends Component {
 
 	/*
-	Example of use hasSidebar and hasBreadcrumb.
+	Example of use hasSidebar and disableAppBreadcrumbs.
 	It must be set in Application.
 	If not set, it is considered as true.
 	
@@ -45,7 +45,7 @@ class Application extends Component {
 	
 	const config = {
 		hasSidebar: false,
-		hasBreadcrumb: false
+		disableAppBreadcrumbs: true
 	}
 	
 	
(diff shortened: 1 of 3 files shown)


ateska (approved):

src/containers/Application.js
```diff
@@ -523,7 +523,7 @@ it is accessible by the sidebar toggler button.
 													name={route.name}
 													render={props => (
 														<>
-															{(this.props.hasBreadcrumb || typeof this.props.hasBreadcrumb === 'undefined') ?
+															{(this.props.hasBreadcrumb || typeof this.props.hasBreadcrumb === 'undefined') && !route.disableBreadcrumbs ?
```
Pe5h4, line 526: Can we unify the terminology here? What I mean is `hasBreadcrumb` might be rewritten to `disableAppBreadcrumbs` and `disableBreadcrumbs` to `disableContainerBreadcrumbs` or something similar, which would have some structural meaning (so when you disable application breadcrumbs, you have to know that you are disabling all the breadcrumbs and that it will be applied also to child components (containers) and when you disable the breadcrumbs in the container, you have to know that you are going to disable breadcrumbs just in the container, nowhere else. I know the logic is implemented, but the naming is kinda confusing to me. What do you think, guys? @ateska @titkov49

titkov49: @Pe5h4 I also was thinking about renaming `hasBreadcrumbs` into `disableAppBreadcrumbs` or something simillar (it also will be a bit about refactoring logic) since breadcrumbs are enabled in apps by default so this will make more sense to me. 
But I'm okay with `disableBreadcrumbs` name since developer disable them in route configuration which I think is understandable that he/she actually disable them only on that route. And also this is written in docs.

titkov49: @ateska Hi Ales,
We discussed with Franta about renaming `hasBreadcrumbs` -> `disableAppBreadcrumbs` and `disableBreadcrumbs`->`disableContainerBreadcrumbs` and we agreed that it should be renamed. But we need your approvement on this. What do you think?

ateska (approved):
Looks ok to me.

Pe5h4 (approved):

PR #196, opened 2021-11-04
Filter those breadcrumbs items which doesn't have names

titkov49:
CHANGES:
- Now items which doesn't have names won't be rendered in breadcrumbs (e.g. library container).
- Also breadcrumbs with only one item will be rendered. (e.g. export container)
- When there no items in breadcrumbs (e.g. when all routes in breadcrumbs doesn't have names) then breadcrumbs container won't be rendered

Screenshot 2021-11-04 at 23 16 55
Screenshot 2021-11-04 at 23 18 58
Screenshot 2021-11-04 at 23 20 37

### DIFF
--- a/src/containers/Breadcrumbs.js
+++ b/src/containers/Breadcrumbs.js
@@ -1,5 +1,5 @@
-import React, { useState, useEffect } from 'react';
-import { useRouteMatch, useLocation, Link } from 'react-router-dom';
+import React from 'react';
+import { Link } from 'react-router-dom';
 
 import { Breadcrumb, BreadcrumbItem } from 'reactstrap';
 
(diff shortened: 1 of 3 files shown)


ateska (approved):

src/containers/Breadcrumbs.js
```diff
@@ -13,25 +13,26 @@ const Breadcrumbs = ({ routes, match }) => {
...
-					</Breadcrumb>
-				</div>
-			}
+			<div className="breadcrumbs">
+				<Breadcrumb>
+					{crumbs.map((crumb, idx) => (
+						<BreadcrumbItem key={idx} active={idx === crumbs.length - 1}>
+							{idx !== crumbs.length - 1 ? 
+								<Link to={crumb.path}>{crumb.name}</Link>
+								: <span>{crumb.name}</span>
```
Pe5h4, line 29: @titkov49 I think we should also add translations for breadcrumbs

Pe5h4 (changes requested):

Pe5h4 (approved):

PR #212, opened 2021-11-23
Feature/update tools module with ability to configure it from zookeeper

Pe5h4:
This PR update Tools module with dynamic configuration option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@
 
 - Add possibility to disable breadcrumbs in specific containers (INDIGO Sprint 211029, [!193](https://github.com/TeskaLabs/asab-webui/pull/193))
 
+- Update Tools module with option to define Tools configuration dynamically (INDIGO Sprint 211112, [!212](https://github.com/TeskaLabs/asab-webui/pull/212))
+
 ### Refactoring
 
 - Renaming configuration option FAKE_USERINFO to MOCK_USERINFO and refactoring code accordingly (INDIGO Sprint 210406, [!91](https://github.com/TeskaLabs/asab-webui/pull/91))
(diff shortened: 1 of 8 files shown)


Pe5h4:
@ateska This is the refactorization of `Tools` module, where option to configure the tools dynamically has been implemented.
- Tools module is possible to configure statically (from config file) or dynamically (in zookeeper with the use of asab-config service). The description how to do so is here https://github.com/TeskaLabs/asab-webui/blob/fcb29b50ef644059bf074e1037a1baa3b81dd45f/doc/tools.md
- At the end, I decided not to move Tools module to BS-WebUI and not to use the BS-WebUI dashboard type of configuration and layout - The reason for it was, that the applications, where we would like to use just the `Tools` module would become rather heavy (since we would have to implement `BS-WebUI` as a submodule to those applications with some of its libraries). Another reason is, that the drag 'n drop dashboard grid used in BS-WebUI Dashboard module has a way broader functionality than we need here for Tools module. Its primary purpose is to drag and drop widgets/items in the grid, the system is more sophsticated and I believe, we dont need to have d'nd functionality in Tools module. Nevertheless the d'nd library itself is also quite heavy.
- So I tried to leave it as simple and functional as it was. Just implemented the option to load the data from Zookeeper via `asab-config` service. Anyway, if there will be any specific demand, to move it to BS-WebUI, I will do that.
tools

ateska (approved):

Pe5h4:
@ateska Tools module updated with the new schema example in the docs file
https://github.com/TeskaLabs/asab-webui/blob/8acf50a02fb39d0c1d87c06f6bfa220136b1c930/doc/tools.md#type-schema-configuration

Also information about option to add `base64` images has been implemented to the docs
https://github.com/TeskaLabs/asab-webui/blob/8acf50a02fb39d0c1d87c06f6bfa220136b1c930/doc/tools.md#images-loaded-as-base64-string

I did the refactorization of the Tools module as per Mithun's asab-config MR on GitLab + I did minor updates in ASAB-Config module.

Broader changes in ASAB-Config module will be done in the next PR as we discussed yesterday.

Here is the Tools flow:
tools-update

PR #218, opened 2021-12-07
Separate user-id translation component

petrKavulok:
translates a single user ID

https://user-images.githubusercontent.com/79644454/145082213-faf29d6f-0cac-4c02-bcff-eb16563a870f.mov

translates multiple user IDs

https://user-images.githubusercontent.com/79644454/145082221-7955f154-8c93-4db6-a0ac-fe4159cbea5d.mov


src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,68 @@
...
+	useEffect(() => {
+		retrieveUserNames();
+	}, [])
+
+    return (
+			<>
+				{usernames && usernames.length != 0 ?
+					Object.keys(usernames).map((key) => {
+						return (
+							<Row>
```
ateska, line 49: It is not a Row ... span or div ... check the DateTime object.

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,68 @@
...
+								<Link className='ml-3' to={{ pathname: `/auth/credentials/${key}`}}>
+									{usernames[key]}
+								</Link>
+							</Row>
+						)
+					}) :
+						id.map((elem) => {
+							return (
+								<Row>
+									<Link className='ml-3' to={{ pathname: `/auth/credentials/${elem}`}}>
```
ateska, line 59: The "id" should be in the hint, so I can hover over the username and it reveals the id in the popup.

Also, consider using icon - in a same way as DateTime uses clock ... it will unify the look&feel.

ateska (changes requested):

ateska:
Also, use the local storage as a cache for translations ... it is quite static operation, so this will save a lot of server requests.

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,163 @@
...
+
+export function UsernameTranslation(props) {
+
+	const id = Array.isArray(props.userId) ? props.userId : [props.userId] ;
+	const apiPath = props.apiPath ?? 'seacat_auth';
+	const { t } = useTranslation();
+	let API = props.app.axiosCreate(apiPath);
+	const [usernames, setUsernames] = useState({});
+	const USERNAMES_CACHE = "usernames";
+	const twoWeeks = 1000 * 60 * 60 * 24 * 14;
```
Pe5h4, line 26: @petrKavulok Can this be a time defined from the properties of the component? What I mean is, we should make it configurable

```
<UsernameTranslation cleanupTime={132456789} ... />
```

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,163 @@
...
+	// test ids
+	// const kryton = 'mongodb:ext:618ced0fe6f6e38ab6a9c8b8';
+	// const rimmer = 'mongodb:ext:618be17866f6ee5de4ae5328';
+	// const id = [kryton,  rimmer];
+
+	const retrieveUserNames = async () => {
+		try {
+			let response = await API.put(`usernames`, id);
+			if (response.data.result !== "OK") {
+				throw new Error(t("TenantDetailContainer|Something went wrong, failed to fetch assigned credentials"));
```
Pe5h4, line 36: @petrKavulok This will definitely not be a `TenantDetailContainer` but something like `ASABUsernameTranslation`

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,163 @@
...
+			} 
+			setUsernames(response.data.data);
+			setUsernamesToCache(id, response.data.data);
+		} catch (e) {
+			console.error(e);
+			props.app.addAlert("warning", t("TenantDetailContainer|Something went wrong, failed to fetch assigned credentials"));
```
Pe5h4, line 42: @petrKavulok The same as mentioned above

src/containers/UsernameTranslation.js
Pe5h4: with some default e.g. 24h

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	const getUsernamesCache = () => {
```
Pe5h4, line 46: @petrKavulok `getUsernamesFromLS` would be better name

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,163 @@
...
+				usernamesCache = JSON.parse(data);
+			}
+		}
+		catch(e){
+			console.error(e.message);
+		}
+		return usernamesCache
+	}
+
+	const setUsernamesToCache = (userId, value) => {
```
Pe5h4, line 63: @petrKavulok `setUsernamesToLS`

Pe5h4 (changes requested):

ateska (changes requested):
Eventually the localstorage should be migrated to this: https://www.w3.org/TR/IndexedDB/

IndexedDB should be better technique than JSON in the LS/

petrKavulok:
Case of a valid ID and an invalid IDs


https://user-images.githubusercontent.com/79644454/147088116-56381e31-9242-484b-bf2c-79aa49817467.mov

ateska:
When resolve fails, don't display "N/A", keep there the original credential_id (be defensive).

ateska:
... and the name of component is `<Credentials />` (it is to diplay credentials ... don't include name of the functionality in the title).

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,150 @@
...
+
+Usage:
+
+import { UsernameTranslation } from 'asab-webui';
+...
+<UsernameTranslation app={props.app} id={['user1-id', 'user2-id', 'user3-id']} apiPath='seacat_auth' cleanupTime={1000 * 60 * 60 * 24 * 14}/>
+
+*/
+
+export function UsernameTranslation(props) {
```
ateska, line 19: `Credentials` ... see my comment.

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,150 @@
...
+			}
+			const usernamesToLS = setUsernamesToLS(response.data.data, userIdsArray, cleanupTime);
+			setUsernames(usernamesToLS);
+		} catch (e) {
+			props.app.addAlert("warning", t("ASABUsernameTranslation|Something went wrong, failed to fetch assigned credentials"));
+		}
+	}
+
+	// compares array of IDs with data in localstorage
+	const matchUserIds = (usernameIDs) => {
```
ateska, line 50: There is no `usernameIDs` or `UserIds` ... there are `credentials_id` and `username`s.

ateska (changes requested):
Naming convention is not aligned with SeaCat Auth.

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,150 @@
...
+			if (data[id]) {
+				item = {
+					id: id,
+					username: data[id],
+				};
+			}
+			if (!data[id]) {
+				item = {
+					id: id,
+					username: 'N/A'
```
ateska, line 133: If not present, use `credentials_id` as username,

ateska (changes requested):

src/containers/UsernameTranslation.js
petrKavulok: i thought `credentials_id` would stand for `'mongodb:ext:1234567890'` and `username` would be `'Ringo'`.
`usernameIDs` on line 50 expects to be an array of strings

src/index.js
```diff
@@ -13,6 +13,6 @@ export { default as SplashScreen } from './containers/SplashScreen';
 export { default as Pagination } from './containers/Pagination';
 export { default as ControlledSwitch } from './containers/ControlledSwitch';
 export { default as UncontrolledSwitch } from './containers/UncontrolledSwitch';
-export { UsernameTranslation } from './containers/UsernameTranslation'
+export { Credentials } from './containers/UsernameTranslation'
```
petrKavulok, line 16: I haven't changed the containers name yet, as the latest changes would not be visible since it would be considered a new file, but it will be changed to `export { Credentials } from './containers/Credentials'` before potential merging

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+		try {
+			ls = JSON.parse(global.localStorage.getItem(name.toString()));
+		} catch (e) {
+			/*Ignore*/
+		}
+	}
+	return ls ? ls : { credentials: [], expiration: new Date().getTime() + cleanupTime };
+}
+
+function setUsernamesToLS(data, userIdsArray, cleanupTime) {
```
Pe5h4, line 114: @petrKavulok `saveUsernamesToLS`

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+function getUsernamesFromLS(name, cleanupTime) {
+	let ls;
+	if (global.localStorage) {
+		try {
+			ls = JSON.parse(global.localStorage.getItem(name.toString()));
+		} catch (e) {
+			/*Ignore*/
+		}
+	}
+	return ls ? ls : { credentials: [], expiration: new Date().getTime() + cleanupTime };
```
Pe5h4, line 111: @petrKavulok Is it necessary to setup expiration with `cleanupTime` to empty credentials? In overall it is kinda strange to set up a cleanupTime in `getUsernamesFromLS`. From my point of view, `cleanupTime` is not necessary for this function. I would expect undefined expiration on empty credentials list or some predefined time, because why would you need to cleanup an empty list/local storage?

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
+import React, { useState, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Link } from "react-router-dom";
+
+/*
```
Pe5h4, line 5: I would expect broader information about this component in the `docs`, not in the code itself. What I mean is this:
```
At the moment works ONLY with apiPath='seacat_auth'
- userId (required, array) - user ids to be translated
- apiPath - default (and only working so far) 'seacat_auth'
- cleanupTime - default set to 24 hours
```
It should be a markdown file, so it should keep the markdown syntax then.

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+	const { t } = useTranslation();
+
+	const apiPath = props.apiPath ?? 'seacat_auth';
+	let API = props.app.axiosCreate(apiPath);
+
+	const userIdsArray = Array.isArray(props.userId) ? props.userId : [props.userId] ;
+
+	const cleanupTime = props.cleanupTime ?? 1000 * 60 * 60 * 24; // 24 hrs
+
+	const [usernames, setUsernames] = useState([]);
```
Pe5h4, line 30: @petrKavulok if the component is named as `Credentials`, it might be also good to rename `usernames` everywhere in the code to `credentials` (also the names of the consts)

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+At the moment works ONLY with apiPath='seacat_auth'
+- userId (required, array) - user ids to be translated
+- apiPath - default (and only working so far) 'seacat_auth'
+- cleanupTime - default set to 24 hours
+
+Usage:
+
+import { Credentials } from 'asab-webui';
+...
+<Credentials app={props.app} id={['user1-id', 'user2-id', 'user3-id']} apiPath='seacat_auth' cleanupTime={1000 * 60 * 60 * 24 * 14}/>
```
Pe5h4, line 15: @petrKavulok You should update the example (there is no `id` if I am not mistaken)

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	// compares array of IDs with data in localstorage
+	const matchUserIds = (credentials_id) => {
+		let currentTime = Date.now();
+		const usernamesInLS = getUsernamesFromLS('Credentials', cleanupTime);
+		let usernamesToRender = [];
+
+		if (usernamesInLS.credentials == undefined || usernamesInLS.credentials.length === 0 || usernamesInLS.expiration <= currentTime) {
```
Pe5h4, line 52: @petrKavulok You can use `Date.now()` directly here, instead of passing it in the variable, which is not used anywhere else (`currentTime`)

```
if (usernamesInLS.credentials == undefined || usernamesInLS.credentials.length === 0 || usernamesInLS.expiration <= Date.now())
```

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+			}
+			const usernamesToLS = setUsernamesToLS(response.data.data, userIdsArray, cleanupTime);
+			setUsernames(usernamesToLS);
+		} catch (e) {
+			props.app.addAlert("warning", t("ASABCredentials|Something went wrong, failed to fetch assigned credentials"));
+		}
+	}
+
+	// compares array of IDs with data in localstorage
+	const matchUserIds = (credentials_id) => {
```
Pe5h4, line 47: @petrKavulok This is kinda confusing me, why is here used `credentials_id` as a props name,
```
const matchUserIds = (credentials_id) =>
```
when it is clearly an array of user IDs? You are passing an array:

https://github.com/TeskaLabs/asab-webui/blob/afe4f174a64b803e78e6cdd7fdcbcedc9ce5eaef/src/containers/UsernameTranslation.js#L69

`credentials_id` should be used for single ID

I suggest to unify the naming, I guess this was some misunderstanding in this conversation: https://github.com/TeskaLabs/asab-webui/pull/218#discussion_r773895258

src/index.js
Pe5h4: I think it would not rewrite the whole file to the point that it would be seen as a new one and the changes would not be visible. Git is clever enough to avoid this.

Pe5h4 (changes requested):

src/containers/UsernameTranslation.js
```diff
@@ -0,0 +1,146 @@
...
+				}
+			}
+			const indexFromLS = dataInLS.credentials.findIndex((itemInLS) => itemInLS.id === item.id);
+			if (indexFromLS === -1) {
+				dataInLS.credentials.push(item);
+			}
+			dataToLS.push(item);
+		})
+		try {
+			global.localStorage.setItem('Credentials', JSON.stringify(dataInLS));
```
titkov49, line 139: You can simplify `global.localStorage.setItem(...)` to `localStorage.setItem(...)`. 
`localStorage` is already a global browser variable.

titkov49: Also I guess that there is no need to `try/catch` localStorage since it's not supported only by browsers which were released 15-20 years ago. Those browsers cannot even start our apps I guess

Screenshot 2021-12-22 at 18 30 58

titkov49: ^^^ Those two comments are appliable to all places where you use localStorage

titkov49 (changes requested):
Otherwise good and approved from my side once those change requests are solved

src/containers/UsernameTranslation.js
ateska: Remark: We don't have "user ids", we have "credentials_ids" ... @byewokko

src/containers/UsernameTranslation.js
ateska: `cleanupTime` should be a globally set variable somewhere else than in the actual container ... it is not container specific, it is feature specific - so it should be handled by some kind of service.

src/containers/UsernameTranslation.js
ateska: * `credentials_id` == 'mongodb:ext:1234567890' ... correct
* `username` == "Ringo" ... correct
* `credentials_ids` should be the array of `credentials_id`

titkov49 (approved):

Pe5h4 (approved):
Looks fine to me

petrKavulok:
@ateska would you please take a look at the changes?

ateska (approved):
Nice.

PR #219, opened 2021-12-08
Update Config module with displaying dynamic values in with patternProperties in schema

Pe5h4:
ASAB - Config refactorisation with patternProperties in schema

### DIFF
--- a/src/modules/asab-config/ConfigContainers/ConfigEditor.js
+++ b/src/modules/asab-config/ConfigContainers/ConfigEditor.js
@@ -422,8 +422,7 @@ export default function ConfigEditor(props) {
 									<div style={{overflow: "auto"}}>
 										<ReactJson
 											src={jsonValues}
-											// TODO: Editing in JSON values is disabled because users can mess with the value types
-											// onEdit={ e => { setJsonValues(e.updated_src)} }
+											onEdit={ e => { setJsonValues(e.updated_src)} }
 											enableClipboard={false}
 											name={false}
 										/>
--- a/src/modules/asab-config/ConfigContainers/ConfigEditor.js
+++ b/src/modules/asab-config/ConfigContainers/ConfigEditor.js
@@ -29,6 +29,7 @@ export default function ConfigEditor(props) {
 	const [ typeData, setTypeData ] = useState(undefined);
 	const [ adHocValues, setAdHocValues ] = useState({});
 	const [ adHocSections, setAdHocSections ] = useState({});
+	const [ configData, setConfigData ] = useState(undefined);
 	const { register, handleSubmit, setValue, getValues, errors, reset } = useForm();
 
 	let App = props.app;
@@ -45,24 +46,61 @@ export default function ConfigEditor(props) {
 
 
 	useEffect(() => {
+		reset({}); // Reset form on config change
 		initialLoad();
-	},[ configType, configName ]); // The container will be re-rendered on configType or configName change
+	}, [ configType, configName ]); // The container will be re-rendered on configType or configName change
 
+	useEffect(() => {
+		if (typeData && configData) {
+			handleConfigValues();
+		}
+	}, [typeData, configData])
+
+	const handleConfigValues = () => {
+		let schema = typeData;
+		let ahValues = {};
+		let ahSections = {};
+		for (var section in configData) {
+			let arrValues = [];
+			let arrSections = [];
+			for (var key in configData[section]) {
+				// Set config values to the schema (if available)
+				setValue('[' + section + '] ' + key, configData[section][key])
+				// Check if config key values are in schema and if not, add it to adhoc values
+				let s = {};
+				let v = {};
+				if (schema.properties) {
+					if (schema.properties[section] == undefined) {
+						s[key] = configData[section][key];
+						arrSections.push(s);
+						ahSections[section] = arrSections;
+					} else if (schema.properties[section]?.properties[key] == undefined) {
+						v[key] = configData[section][key];
+						arrValues.push(v);
+						ahValues[section] = arrValues;
+					}
+				}
+			}
+		}
+
+		// TODO: HANDLE ad hoc values for patternIndexes
+		setAdHocValues(ahValues);
+		setAdHocSections(ahSections);
+		setJsonValues(configData);
+	}
 
 	const initialLoad = async () => {
 		let values = undefined;
-		let type = undefined;
+		let schema = undefined;
 		setConfigNotExist(false);
 
 		try {
 			let response = await ASABConfigAPI.get("/type/" + configType);
 			// TODO: validate responses which are not 200
-			type = response.data;
-			if (type.result == 'FAIL') {
+			schema = response.data;
+			if (schema.result == 'FAIL') {
 				App.addAlert("warning", t(`ASABConfig|Something went wrong! Unable to get data`, {type: configType}));
 				return;
-			} else {
-				setTypeData(type);
 			}
 		}
 		catch(e) {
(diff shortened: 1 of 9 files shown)


Pe5h4:
@ateska This is the first iteration of refactorization of ASAB-Config module based on friday's discussion and outputs

If there will be anything broader to add/refactor from you or something unclear, we can have a short call about it.

Values to render and section (pattern)properties are stored in formStruct in following flow described in points:

After data from endpoints are obtained:
 1] add sections and section keys from properties to formStruct with appropriate data (if properties present in the schema)
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L109
 https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L115
 2] add sections and section keys from patternProperties to formStruct with appropriate data (if patternProperties sections are present in the schema)
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L144
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L148

https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L176-L181

 3] add rest of the values to adHoc
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L171-L174
 4] set values from formStruct and from adHoc for register (with the usage of `setValue` comonent of react-hook-form)
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L186-L201
 5] render input forms and data according to formStruct properties
https://github.com/TeskaLabs/asab-webui/blob/a8d7e819d99300392a405b53ad36528bb3634841/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L307-L328

Sections with adHoc cannot be saved now (it does not fit the schema, service is throwing errors) and support for patternProperties will be added as well (now we cannot save values to configs with patternProperties) - I discussed this with Mithun and he is going to work on that next sprint.

Here is the flow with examples of configuration with adHoc sections or/and adHoc keys (values). On gif image there is also an example of schemas with patternProperties (and as well with adHocs):

config-refactorisation

I have also question - should we support nested patternProperties (meant for keys)? I suppose for our `section-key-value` configs this wont be necessary. Here is the example:

```
...
 "patternProperties": {
        "^section:source:(?<varname>.*)$": {
            "type": "object",
            "title": "Section configuration",
            "description": "Secition source",
            "default": {},
            "examples": [
                {
                    "^section:source:name:(?<keyname>.*)$": "My Key"
                }
            ],
            "required": [
                "^section:source:name:(?<keyname>.*)$"
            ],
            "patternProperties": {
                "^section:source:name:(?<keyname>.*)$": {
                    "type": "string",
                    "title": "Name",
                    "description": "Fill the name of the Key",
                    "default": "",
                    "examples": [
                        "MyOwnKey"
                    ]
                }
            },
            "additionalProperties": false
        },
...
```

ateska:
Support for the `pattern on key` will be added eventually (if ever).

src/modules/asab-config/ConfigContainers/ConfigEditor.js
```diff
@@ -22,113 +22,181 @@ import {
...
-					let v = {};
-					if (type?.properties[section] == undefined) {
-						s[key] = values[section][key];
-						arrSections.push(s);
-						ahSections[section] = arrSections;
-					} else if (type?.properties[section]?.properties[key] == undefined) {
-						v[key] = values[section][key];
-						arrValues.push(v);
-						ahValues[section] = arrValues;
+		let fs = {};
```
ateska, line 100: Here we need more descriptive variable name.

src/modules/asab-config/ConfigContainers/ConfigEditor.js
```diff
@@ -22,113 +22,181 @@ import {
...
+		let schemaProps = {};
+		let data = {};
```
ateska, line 102: Here we need more descriptive variable name, `data` doesn't reveal too much.

ateska (changes requested):
Few remarks but this is a brilliant direction.

Pe5h4:
@ateska Updates:
- variable names for formStruct renamed https://github.com/TeskaLabs/asab-webui/blob/4ab015a9c5c620246a068dcb7ce2746f569c6b9d/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L98-L100
- feature to handle cases, when key-value is not present in the configuration, but schema is present for this case (it can be the case of an empty config) https://github.com/TeskaLabs/asab-webui/blob/4ab015a9c5c620246a068dcb7ce2746f569c6b9d/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L118-L122 - empty string is added to the section-key (we cannot use undefined value since it would not be rendered properly, but we can use empty string for every input type [number, boolean, etc.])
- implemented function to convert values based on value types on submit. onSubmit function converts all the updated values to string, so it has to be parsed back by the appropriate `type` before data are being processed to the service (as we discussed yesterday with you and Mithun) https://github.com/TeskaLabs/asab-webui/blob/4ab015a9c5c620246a068dcb7ce2746f569c6b9d/src/modules/asab-config/ConfigContainers/ConfigEditor.js#L313
- I commented editing values from JSON viewer (so now it is disabled), since you can specify the type of the value from JSON viewer and if you choose the wrong one, the configuration will not be saved. Anyway from what I remember, we were speaking of removing the JSON view completely in the future.
- other minor updates as seen from the commits

When Mithun will provide functional update on asab-config service, I will add a new gif image with the flow

ateska:
Looks good to me so far.

Pe5h4:
@ateska In addition to above updates you have already aware of, I extended documentation for asab-config module (schema and config examples, supported input types) https://github.com/TeskaLabs/asab-webui/blob/da080a3fb044fdfe07dd33b922b36be40e305ad4/doc/asab-config.md

The creation of new (empty) configuration from asab-config module and sorting sections in the tree menu by name (as we spoke of) will be done in the next iteration

Here is the read/save flow for properties and pattern properties also with ad hoc values and ad hoc sections:
pattern-props-asab-config

ateska:
Why there is `config` in  `/config/Discover` path in the URL? ... that seems to be dangerous to me.

ateska:
Also word "key" in the config item is a bit redundant already.

ateska:
Tenant value in a config needs to be removed.

Let's have a review workshop for this, there are few additional remarks I have.

ateska (approved):

Pe5h4:
> Why there is `config` in `/config/Discover` path in the URL? ... that seems to be dangerous to me.

@ateska This is a path to the config section. We decided to have path URL as `config/sectionName/configName`, (default is `config/$/$`) so the `config/Discover` just points to the section.

Pe5h4:
> Also word "key" in the config item is a bit redundant already.

@ateska Yes, this is for the purpose of exercise. The name or the description is taken from the schema. Anyway, we should have a workshop about how to properly configure schemas (I suppose also together with Mithun)

Anyway, lets have a talk about it on the review, I will merge it only if there will be only after the review and only if there will be minors, which can be done in the next iteration.

PR #220, opened 2022-01-04
moment sets its locales based on user's browser's language and displays date accordingly

petrKavulok:
Snímek obrazovky 2022-01-04 v 17 02 55
Snímek obrazovky 2022-01-04 v 17 02 20
Snímek obrazovky 2022-01-04 v 17 01 48

### DIFF
--- a/demo/conf/config.js
+++ b/demo/conf/config.js
@@ -1,8 +1,6 @@
 module.exports = {
 	app: {
-		API_PATH: 'api',
-		modules: ["HomeModule"],
-		locales: ["cs"]
+		modules: ["HomeModule"]
 	},
 	webpackDevServer: {
 		port: 3000,
--- a/demo/conf/config.js
+++ b/demo/conf/config.js
@@ -2,7 +2,7 @@ module.exports = {
 	app: {
 		API_PATH: 'api',
 		modules: ["HomeModule"],
-		momentLocales: ["cs", "en-gb"]
+		locales: ["cs"]
 	},
 	webpackDevServer: {
 		port: 3000,
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -34,7 +34,7 @@ import FooterService from '../services/FooterService';
 import TenantSelectionCard from '../modules/tenant/selector/TenantSelectionCard';
 
 import { ADD_ALERT, SET_ADVANCED_MODE, CHANGE_HELP_URL } from '../actions';
-import LanguageReducer, { CHANGE_LANGUAGE } from '../reducers/LanguageReducer';
+import { CHANGE_LANGUAGE } from '../reducers/LanguageReducer';
 
 
 class Application extends Component {
--- a/src/containers/Application.js
+++ b/src/containers/Application.js
@@ -24,6 +24,8 @@ import alertsReducer from '../alerts/reducer';
 import helpButtonReducer from '../helpButton/reducer';
 import sidebarReducer from '../sidebar/reducer';
 
+import languageReducer from '../reducers/languageReducer'
+
 import ReduxService from '../services/ReduxService';
 import ConfigService from '../config/ConfigService';
 import HeaderService from '../services/HeaderService';
@@ -32,6 +34,7 @@ import FooterService from '../services/FooterService';
 import TenantSelectionCard from '../modules/tenant/selector/TenantSelectionCard';
 
 import { ADD_ALERT, SET_ADVANCED_MODE, CHANGE_HELP_URL } from '../actions';
+import LanguageReducer, { CHANGE_LANGUAGE } from '../reducers/LanguageReducer';
 
 
 class Application extends Component {
@@ -97,6 +100,7 @@ class Application extends Component {
 		this.ReduxService.addReducer("advmode", AdvancedModeReducer);
 		this.ReduxService.addReducer("helpButton", helpButtonReducer);
 		this.ReduxService.addReducer("sidebar", sidebarReducer);
+		this.ReduxService.addReducer("language", languageReducer)
 
 		this.DefaultPath = props.defaultpath;
 
@@ -470,6 +474,13 @@ class Application extends Component {
 				})
 			}
 		}, [])
+	}
+
+	changeLanguage(lang) {
+		this.Store.dispatch({
+			type: CHANGE_LANGUAGE,
+			language: lang
+		});
 	}
 
 
--- a/src/containers/DateTime.js
+++ b/src/containers/DateTime.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import moment from 'moment/min/moment-with-locales';
+import moment from 'moment';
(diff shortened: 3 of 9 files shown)


ateska:
Please check the increase of size of the package as discussed.

ateska (changes requested):
As said above, please provide metrics on how this enlarges the app package.

src/containers/DateTime.js
```diff
@@ -37,6 +37,7 @@ export function DateTime({ value, format }) {
 	}
 
 	const m = isNaN(value) || value > 9999999999 ? moment(value) : moment(value * 1000);
+	m.locale(window.navigator.language.slice(0, 2));
```
ateska, line 40: The locale should be taken from the current language settings.

ateska (changes requested):

src/containers/DateTime.js
Pe5h4: @ateska We had a discussion about this with @petrKavulok , so does it mean that it should follow the language of the application (taken from i18n) and not the browser's language?

ateska: Yes, it sounds to me logical that then I switch to En, I see english names on moths etc.

Also, the language is set initially based on the browser lang + possibilities of the app (aka available localizations), which is also "correct".

petrKavulok:
> As said above, please provide metrics on how this enlarges the app package.

Snímek obrazovky 2022-01-05 v 11 11 32

ateska:
Can this be loaded using lazy loader? @titkov49

src/containers/DateTime.js
petrKavulok: @ateska just to be clearify if I understood correctly - 
- initially the language should be set by the browser language
- then if user changes the apps language, the date should respect it and adjust accordingly

is that correct?..

ateska: From the perspective of this component, you should always respect the language selection visible in the top bar - which is the language used for localization.

ateska: https://github.com/TeskaLabs/asab-webui/blob/b0487a8c7b082c7b77b58dcb407f8b587f4f2490/src/modules/i18n/dropdown.js#L31

`i18n.language` ... specifically

BTW - that language should be in the reduct store so that you don't introduce dependency on i18n.

... and if (and only if) the language is not found in the redux store, then use the language from the browser.

titkov49:
> Can this be loaded using lazy loader? @titkov49
@ateska 
By `this` you mean component or only this `moment-with-locales.js`? If first then I guess we can lazy-load it if we wrap DateTime into `Suspense` first

ateska:
Yes, that's what I meant.

titkov49:
@ateska @fpesek @petrKavulok
English locale is default one and included into moment library (cannot be lazy-loaded). Other locales are now lazy-loaded for moment only once when switched to another language (it's presented in the video with czech language).
https://user-images.githubusercontent.com/37152497/150449179-dcd7bb62-22f2-4983-906f-33d93e5d7046.mov

titkov49:
@ateska @petrKavulok @Pe5h4
Hi,
After implementation of lazy-loading of locales (https://github.com/TeskaLabs/asab-webui/pull/220/commits/3ea50c7140e026b59b364cdf04587d00106a22ed) I was checking bundle size with webpack-bundle-analyzer and found out that those locales that we don't use are already removed during building process and suddenly I remembered that we've already solved that issue in this PR https://github.com/TeskaLabs/asab-webui/pull/98 actually 
I suppose that removing all unused locales during building process is a better solution than lazy-load those locales each time we wanna localize moment. So the solution of reducing moment's locales pack is that we just shouldn't import `moment/min/moment-with-locales`
Also I removed `en-gb` default locale since I found out that `en` locale is already included into moment package by default.

And I also attach screenshots of moment package checked with webpack-bundle-analyzer plugin:
Screenshot 2022-01-21 at 3 20 14
Screenshot 2022-01-21 at 3 20 20

titkov49:
And also video of switching between languages 

https://user-images.githubusercontent.com/37152497/150455079-b00de686-ee10-40f7-84d5-3e67c20fa55e.mov

src/containers/DateTime.js
```diff
@@ -28,20 +29,30 @@ The default format is `lll` -> `Aug 22, 2020 1:13 PM`
...
 	)
 }
+
+// function mapStateToProps(state) {
+//     return {
+//         language: state.language
+//     }
+// }
+
+// export default connect(mapStateToProps)(DateTime)
```
titkov49, line 58: I suppose this should be removed

src/containers/Application.js
```diff
@@ -32,6 +34,7 @@ import FooterService from '../services/FooterService';
 import TenantSelectionCard from '../modules/tenant/selector/TenantSelectionCard';
 
 import { ADD_ALERT, SET_ADVANCED_MODE, CHANGE_HELP_URL } from '../actions';
+import LanguageReducer, { CHANGE_LANGUAGE } from '../reducers/LanguageReducer';
```
titkov49, line 37: As I see you should also remove LanguageReducer from this import

src/reducers/languageReducer.js
```diff
@@ -0,0 +1,16 @@
...
+
+
+export default (state = initialValue, action) => {
+    switch (action.type) {
+        case CHANGE_LANGUAGE:
+            return { ...state, language: action.language }
+        default:
+            return state
+    }
+}
```
titkov49, line 16: Add new line please

titkov49 (changes requested):

demo/conf/config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
 	app: {
 		API_PATH: 'api',
```
Pe5h4, line 3: @petrKavulok Can you also please remove this line, which is kinda obsolete? Thanks

src/modules/i18n/dropdown.js
```diff
@@ -20,6 +22,16 @@ function LanguageDropdown(props) {
 		});
 	}
 
+	const dispatch = useDispatch();
+
+	useEffect(() => {
```
Pe5h4, line 27: @petrKavulok I suppose, you can remove this useEffect and just the useEffect on lines 31-33 is enough and should stay, since it will be triggered anyway (but test that please)
https://github.com/TeskaLabs/asab-webui/blob/b0c4090160a7462bc8a34521644d900d9963fd64/src/modules/i18n/dropdown.js#L31-L33

Pe5h4 (approved):

Pe5h4:
@petrKavulok Can we have some new video with the new changes?

petrKavulok:
https://user-images.githubusercontent.com/79644454/152391257-859f8e1b-b41a-4703-8788-e499225ff763.mov

ateska (approved):
Nice.

PR #227, opened 2022-01-17
Draft: Dark Mode (Dark Theme)

titkov49:


titkov49:
@ateska
Also AboutModule was refactored base on newer changes which were done in cards (cards on screenshots have only fields which exist for locale dev, for builds there are more fields):
Screenshot 2022-05-04 at 15 41 26

Screenshot 2022-05-04 at 15 40 52

src/components/DataTable/Table.js
```diff
@@ -124,7 +127,7 @@ const Headers = ({ headers, advmode }) => (
 			)}
 		</colgroup>
 
-		<thead className="thead-light data-table-thead">
+		<thead className={`thead-${theme === "theme-dark" || !theme ? "light" : "dark"} data-table-thead`}>
```
ateska, line 130: Use the CSS selector, please.

src/components/DataTable/table.scss
```diff
@@ -0,0 +1,98 @@
+@import '../../styles/index.scss';
+
+// STYLES FOR Table.js
+
+.table .thead-light th, .table .thead-dark th {
```
ateska, line 5: Try to avoid use of the themes in the CSS.
If avoiding is not possible, then use root selector.

ateska (changes requested):

src/components/DataTable/table.scss
titkov49: @petrKavulok ^^^

src/components/DataTable/Table.js
titkov49: @petrKavulok ^^^

src/styles/index.scss
```diff
@@ -36,9 +24,25 @@ html {
 
 		color: $text-color;
 		background-color: $body-bg;
+
+		p, span, h1, h2, h3, h4, h5, h6, div {
+			color: $text-color;
+		}
```
titkov49, line 30: @ateska
Hi, rewriting those elements here is needed b/c otherwise they may inherit text color not from body, but from some another parent container which is closer to element. E.g. `Table` from reactstrap adds to `tr` element some black font-color on hover which overrides our color from body. In order not to trace all occasions where we can find such behaviour and fix them explicitly, we decided to assign our text color to all of the elements. Such approach ofc may lead to problem like we have with `react-json-view` when our styles rewrites styles of some component which also rely on inheritance and when we don't want to rewrite those. But in case of `react-json-view` that's actually a problem of `react-json-view` itself. And in our applications we've found only one library which has such problem. BTW solution for that one we have below (lines 32-37)

ateska: Ok. Understand.

ateska:
There is a conflict - please resolve.

Pe5h4:
@petrKavulok @Aringoaway Is this PR duplicate of https://github.com/TeskaLabs/asab-webui/pull/282 and https://github.com/TeskaLabs/asab-webui/pull/286 ? If so, can we close it? If not, what are the missing parts in  https://github.com/TeskaLabs/asab-webui/pull/282 and https://github.com/TeskaLabs/asab-webui/pull/286 ?

PR #229, opened 2022-01-31
Implement create and remove config file feature to ASAB-Config module 

Pe5h4:
- Add feature to create new config file in the section
- Add feature to remove the config file in the section


Pe5h4:
@ateska In this draft PR of ASAB WebUI, I have updated ASAB-Config module with option to create/remove a configuration. There has been some changes since review (data table has been implemented for every section, where user has an overview on the section and can manage the configuration from there). Anyway, I would appreciate a workshop about the implemented features, its look and most importantly - how to handle schemas with pattern properties, when creating a new configuration (as we spoke on review, for this, we would have to think of creating a specific schemas)

Here is the current look

1] creating / removing configuration in section with `schema properties`
create_and_delete_config_properties

2] creating / removing configuration in section with `schema pattern properties`
create_and_delete_config_patternproperties

ateska:
Yes!

ateska (approved):

ateska (changes requested):
Looks good ... but I suggest to use different way of "Add" button.

"Add" itself should be a dropdown with selection of what to add  ... that seems to me more usable and logical.

Pe5h4:
@ateska Refactored. The dropdown button is disabled for configuration files with no pattern props
config-dropdown

ateska (approved):

PR #231, opened 2022-02-07
Library module

titkov49:
Move `Library` module from `lmio-webui` to `asab-webui`. 

Other changes: 
- Now only tree menu is reloaded (data is refetched) when file is switched to/from disabled instead of reloading whole screen. Video example is in comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,8 @@
 
 - Fix networking indicator progress bar position on the screen (INDIGO Sprint 220204, [!234](https://github.com/TeskaLabs/asab-webui/pull/234))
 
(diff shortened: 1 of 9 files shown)


titkov49:
https://user-images.githubusercontent.com/37152497/152789301-add6d0e3-ccb7-4b75-b9ca-df38ae7bf90f.mov

ateska (approved):
Yes, this is exactly what we need.

Pe5h4 (changes requested):
@titkov49 Can you please write some small documentation on Library (how to use it) and add to the doc file also locales used in the Library container? Thanks! Otherwise looks great!

petrKavulok (approved):

Pe5h4 (approved):

PR #232, opened 2022-02-07
Restructure src folder

titkov49:
**Changes:**
1. `components` folder was created and most of reusable components (those components which we export from asab-webui) were moved there:
- ControlledSwitch
- UncontrolledSwitch
- DataTable
- KnowledgeBase
- ContentLoader
- Credentials
- DateTime
- Humanize
- Pagination
- Spinner

2. `HelpButton` component and it's reducer was moved to `Header` folder since they are used only there.

3. `Alert` folder was created inside `containers`. `AlertsComponent` component and it's reducer were moved to `Alerts` folder. New path for alerts is `src/containers/Alerts`. Also `AlertComponent` was renamed into `index.js`

4. Sidebar reducer was moved to `Sidebar` folder. New path for sidebar reducer is `src/containers/Sidebar/reducer`

5. `LoadingSpinner.js` was removed. It is obsolete component which was used as part of `DataTable.js`

**Questions**:
1. What should I do with `Duration.js` component (`src/containers/Duration.js`)? It seems that it's not used anywhere now.


CHANGELOG.md
```diff
@@ -196,6 +196,8 @@
 
 - Refactor `Tools` module with different Tools configuration processing based on ASAB-Config service changes (INDIGO Sprint 220107, [!222](https://github.com/TeskaLabs/asab-webui/pull/222))
 
+- Move all reducers to reducers folder (INDIGO Sprint 220204, [!232](https://github.com/TeskaLabs/asab-webui/pull/232))
```
Pe5h4, line 199: @titkov49 Technically its not all reducers, but only application reducers, since there are also reducers in particular modules of asab-ui

Pe5h4 (approved):
Good

CHANGELOG.md
titkov49: Yeah, I meant all reducers which are loaded in Application.js directly. Others are part of modules and should stay on their places

ateska:
This is not correct. I want to locate components based their "business topic" (alert) not technical classification (reducer).
A reason for it is that it allows smoother decomposition into smaller projects if some part grows "too much".

titkov49:
@ateska @Pe5h4 re-review please

Pe5h4 (approved):
@titkov49 In overall it looks good from my point of view, the structure is more clear now. `src/containers/Duration.js` can be removed.

ateska (approved):

PR #236, opened 2022-02-17
Replace moment with date-fns and implement lazy-loading for locales

titkov49:
Other changes:
1) Output bundled js file was splitted into two files + files for date-fns (including locales): one contains our source code and the second one contains node modules (improves loading time b/c of caching node_modules and date-fns modules)
2) `chunkhash` in the name of files was replaced with `contenthash`. Now new hash for files will be generated base on changes in files' content
3) Outdated `extract-text-webpack-plugin` was replaced with newer `mini-css-extract-plugin`
4) `postcss` preprocessor was added with plugins `css-mqpacker` (compose styles), `autoprefixer` (add prefixes for browsers compatibility) and `cssnano` (minimize and optimize output css). Also `postcss.config.js` file was created in root in order to enable plugins
5) Added basic source maping
6) Babel presets bug was fixed (was `@babel/env`, now `@babel/preset-env`)
7) Node modules was excluded from babel compiling process (those are already compiled)
8) `locales` from configuration was removed. Now date localization is aligned with i18n


titkov49:
Screenshot 2022-02-17 at 14 52 55

ateska:
And what is the absolute size?

src/containers/DateTime.js
```diff
@@ -39,14 +40,33 @@ export function DateTime(props) {
...
+		const importedLocale = await import(
+			/* webpackMode: "lazy", webpackChunkName: "df-[index]" */
+			`date-fns/locale/${language}/index.js`
+		);
+		setLocale(importedLocale.default);
+	}	
+
+
+	const date = isNaN(props.value) ? format(parseISO(props.value), 'PPp', { locale: locale }) :
+		props.value > 9999999999 ? format(props.value, 'PPp', { locale: locale }) :
```
Pe5h4, line 63: @titkov49 @ateska What do you think about using `Pp` as a time format? To me it looks a bit more convenient, when there are just numbers as a date. https://date-fns.org/v2.28.0/docs/format

Pe5h4 (approved):

src/containers/DateTime.js
ateska: Ok, let's try that and we will quickly see.

PR #237, opened 2022-02-18
DataTable custom description

titkov49:
Also max char limit was removed

### DIFF
--- a/src/containers/DataTable/index.js
+++ b/src/containers/DataTable/index.js
@@ -29,7 +29,7 @@ export function DataTable ({
 	buttonWithAuthz, sort, noItemsComponent,
 	customButton, customComponent,
 	customRowStyle, customRowClassName,
-	customDescription,
+	customCardBodyComponent,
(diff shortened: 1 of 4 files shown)


ateska (approved):

src/containers/DataTable/index.js
```diff
@@ -29,6 +29,7 @@ export function DataTable ({
 	buttonWithAuthz, sort, noItemsComponent,
 	customButton, customComponent,
 	customRowStyle, customRowClassName,
+	customDescription,
```
Pe5h4, line 32: @titkov49 Maybe it would be better to call it `customCardBodyComponent` ? Or I suppose its a component, right? So developer can add not only a text component( a paragraph for instance) but also a button or something else, right?

Also I miss a documentation about how to use it

Pe5h4 (changes requested):

Pe5h4 (approved):

PR #239, opened 2022-02-21
DateTime fix return condition

titkov49:
Other changes:
- Remove `style-loader`. `MiniCssExtractPlugin.loader` is adding style element to html instead.
- Add title with distance from date to DateTime
- Remove `package.json` from `asab-webui` root
- Add defense from invalid dates


titkov49:
@Pe5h4 
https://user-images.githubusercontent.com/37152497/155122002-64533df3-05d7-4471-9e09-8f4ac3f36991.mov

src/containers/DateTime.js
```diff
@@ -30,23 +30,54 @@ The default format is `lll` -> `Aug 22, 2020 1:13 PM`
...
+		);
 	}
 
-	const m = isNaN(props.value) || props.value > 9999999999 ? moment(props.value) : moment(props.value * 1000);
-	m.locale(language.slice(0, 2));
+	if (new Date(props.value).toString() === "Invalid Date") {
+		return (
+			<span className='datetime'>
+				<i className="cil-clock pr-1"></i>
+				Invalid Date
```
Pe5h4, line 64: @titkov49 Is it possible to translate it?

Pe5h4 (approved):
Looks cool. Just one question mentioned above - if it would be possible, please update

src/containers/DateTime.js
titkov49: > @titkov49 Is it possible to translate it?

@Pe5h4 I can edit it as common translation:
```js
t("Invalid Date")
```

PR #241, opened 2022-02-23
Refactor Microservices API calls

titkov49:


src/modules/asab-config/MicroservicesContainers/MicroservicesContainer.js
```diff
@@ -36,19 +36,23 @@ export default (props) => {
...
-				props.app.addAlert("warning", t('Failed fetch'));
-			})
+		getMicroservicesList();
 	}, [page, limit]);
 
+	const getMicroservicesList = async () => {
+		try {
+			const response = await ASABConfigAPI.get('/microservices', { params: { p: page, l: limit }});
+
+			if (response.data.result !== "OK") throw new Error(res);
```
Pe5h4, line 46: @titkov49 This `throw new Error(res)` does not seem good to me, `res` is not defined anywhere

Pe5h4 (changes requested):

src/modules/asab-config/MicroservicesContainers/MicroservicesContainer.js
titkov49: @Pe5h4 Oh yeah, you're right. Since it's in `try` this also throws error which is catched so I didn't even noticed that

petrKavulok (approved):

Pe5h4 (approved):

PR #243, opened 2022-02-26
Completely remove moment  - replace moment with date-fns in DataTable 

titkov49:


src/containers/DataTable/index.js
```diff
@@ -63,7 +63,8 @@ export function DataTable ({
 				return JSON.stringify(item[header.key])?.replace(',', ';');
 			}).join(',')).join('\n');
 		let blob = new Blob([csv], {type: "text/csv;charset=utf-8"});
-		saveAs(blob, `${title.text.replace(' ', '_')}_${m().format('D-MM-YYYY')}.csv`);
+		let name = title?.text || "";
+		saveAs(blob, `${name.replace(' ', '_')}_${format(new Date, 'dd-MM-yyyy')}.csv`);
```
Pe5h4, line 67: @titkov49 Do we actually use `downloadHandler` anywhere? Since it can download only csv, I dont think it have to be present in DataTable component especially when you can specify Download as a custom component of DataTable

Then maybe we dont need `date-fns` at all in DataTable

Pe5h4 (changes requested):

src/containers/DataTable/index.js
titkov49: @Pe5h4 I don't remember if this is used anywhere but it was a request from Ales to have such functionality. And I don't see any reasons to remove that when it's already there

Pe5h4: @titkov49 Ok I see, but in any case, there is no reason to user `date-fns` here for formatting, since it can be done purely using JS Date component

```
let date = new Date();
...
saveAs(blob, `${name.replace(' ', '_')}_${date.getDate()}-${date.getMonth()}-${date.getFullYear()}.csv`);
```

titkov49: @Pe5h4 Yeah, but there is no reason not to use date-fns since we already use that in our application and it will not increase application size anymore

Pe5h4: ok

Pe5h4 (approved):

ateska (approved):

PR #245, opened 2022-02-28
Add option to dynamically remove sidebar items

Pe5h4:
- Add option to dynamically remove sidebar items via asab-config service and zookeeper


Pe5h4:
@ateska In this PR, feature to dynamically hide Sidebar items of the Application has been implemented.

Sidebar items are fully loaded everytime when asab-config service is broken, wrongly configured, unavailable or configuration in zookeeper is not set properly.

Documentation has been extended to help people to setup zookeeper environment (there is also schema example):
https://github.com/TeskaLabs/asab-webui/blob/01e7a61f8c3a0d169f188961ec9acc30d18dc27b/doc/sidebar.md#dynamic-sidebar-configuration

There are 2 questions I have from my side:

1] There is a small glitch, that when application is loaded/re-loaded with some hidden sidebar items, the items to be hidden appear in a sidebar for a milisecond. This is caused by the fact of asab-config service with zookeeper is rather slow and I did not wanted to extend the time of application load by putting the request for asab-config service to init of `Application.js` file and to extend the loading time of the app. I hope it is alright, since most of the users will not notice that - and also - its just a kind of softcheck. On the opposite way, I can refactor it to the point, that items in the sidebar is going to be rendered only after the configuration is loaded from asab-config service / or fully loaded when the request to a service will fail (due to e.g. unavailable service). Should I refactor it, or is it alright? You can take a look to attached gif (the glitch will happen when the item will be set to stay hidden and then the page is refreshed. This action is too fast, that its not even captured in the gif)
dynamic-sidebar


2] Should I leave this message as a warning in the console of the browser, when e.g. asab-config service is unavailable / Sidebar section has not been set in Zookeeper, etc. ?
Selection_343

ateska:
@Pe5h4 

Q1: It is OK. For now - I guess - I expect that the loading time of the app will grow eventually and this may become a visible issue. Then we can address that.

Q2: Console warning is OK.

Pe5h4:
@ateska @titkov49 I resolved the glitch by adding a state which return `true` in `useConstructor` custom hook for `readyToRender` const, whenever the page is reloaded. You can check it here: https://github.com/TeskaLabs/asab-webui/pull/245/commits/f719946c94fbafd0b5ee3a74feca83ee69fbc336

It is a variation on Dima's suggestion

ateska (approved):

titkov49 (approved):

Pe5h4:
@ateska @titkov49 Based on review's discussion, I created Sidebar service to handle the sidebar items to be hidden in the sidebar. This is being done during the application initialization. So because of this, I got rid of the glitch/delay where the sidebar item which has been configured as hidden has been rendered and then removed when loading the sidebar items when configured dynamically. It is (by my opinion) more reliable and effective, than it was presented on the review.

The process is now hidden behind app initialization splashscreen.

When the ASAB-Config service, which is responsible for loading the items to be hidden for sidebar, is unavailable or broken, there is no delay in the splashscreen and all the sidebar items are rendered straight away.

ateska (approved):

PR #248, opened 2022-03-03
Implement option to remove section from config

Pe5h4:
- Implement option to remove config section in ASAB Config module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,9 @@
 
 - Add option to dynamically hide sidebar items via ASAB-Config service (INDIGO Sprint 220218, [!245](https://github.com/TeskaLabs/asab-webui/pull/245))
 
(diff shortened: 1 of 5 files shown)


Pe5h4:
@ateska In this ASAB-Config PR, an option to remove Section from Configuration has been implemented. There has been some changes based on review's remarks aswell

- button for section removal is now outlined
- button for section removal and section addition is only visible for configurations with pattern properties
- the sections for `Add` button dropdown are taken from configuration schema (when user add the new section, section key is generated based on count of the selected section pattern in the config, but if there should be any clash with the same key name, there is a condition implemented, which will add random letter/number behind the new section name and the check runs again)
- duplicate code has been unified

Here is how it looks like in test env
asab-config-remove-section

ateska (approved):

PR #253, opened 2022-03-22
Timestamp to string component

petrKavulok:
- timeToString() function is used to translate time value to time and date format based on locales.
- returns string (unlike DateTime)


petrKavulok:
https://user-images.githubusercontent.com/79644454/159685888-98a09f0a-9d38-4443-babe-493dc11cc0ce.mp4

src/components/timeToString.js
```diff
@@ -0,0 +1,54 @@
...
+export const timeToString = (value) => {
+    const isMounted = useRef(null);
+	const [locale, setLocale] = useState('');
+	const language = useSelector(state => state.language.language)
+	
+	// Sets isMounted indicator to true
+	// when component has been mounted (componentDidMount analogue)
+	// and sets isMounted to false before component 
+	// is unmounted (componentWillUnmount analogue)
+	useEffect(() => {
```
titkov49, line 14: This function shouldn't use any react hooks b/c in that case it can be used only as React Component.
We need only value, not a component.

titkov49 (changes requested):

src/components/timeToString.js
Pe5h4: I second that

Pe5h4:
@petrKavulok Can we test that on BS-Charts like last time? I suppose, now it should work there as well since you got rid of state components.

petrKavulok:
@Pe5h4 it does now

https://user-images.githubusercontent.com/79644454/161523282-828a0642-7575-4c7a-9044-f24d5ddf92a3.mp4

titkov49 (approved):

Pe5h4 (changes requested):
@petrKavulok Can you please write a documentation, how to use it when someone would like to return only a string (with translation possibly)?

doc/timeToString.md
```diff
@@ -0,0 +1,25 @@
...
+
+```javascript
+import { timeToString, useDateFNSLocale } from 'asab-webui';
+
+const MyComponent = () => {
+    const lang = useDateFNSLocale();
+
+    return (
+        <>
+            Current date and time is {timeToString(new Date(), lang)}
```
Pe5h4, line 18: @petrKavulok Well, for this case, I may use the original component, since its in the render part. Can you please add the documentation, how to use timeToString also outside of render (e.g. as a value return)?

Or do it different way - create one doc about DateTime component, and inside of it, add sections with examples how to use the original `DateTime` component in render, and how to use timeToString method.

Pe5h4 (changes requested):

Pe5h4 (approved):

PR #254, opened 2022-03-30
Add dropdown button refactorisation

Pe5h4:
- Refactor add button dropdown in asab-config for pattern props that it displays the title of the section defined in schema


Pe5h4:
@ateska This PR is a group of 3 changes made in ASAB-UI

1] ASAB-Config Module - for Add section dropdown button in configuration with pattern properties, there used to be programatical section names as. e.g  `sidebar:item` in the dropdown list. This has been replaced with the `title` of the section taken from the schema, e.g `Sidebar item`, so now it is human readable. The fallback is programatical section name.


asab-config-replace-name-in-add-dropdown

2] As per discussion on Slack, I have extended condition in Sidebar service, which now check for presence of ASABConfigModule to trigger a request for configuration of sidebar. If ASABConfigModule is not present (loaded) in the Application itself, the call is not triggered, so we will avoid requests to asab-config service in deployments, where asab-config service is not needed. https://github.com/TeskaLabs/asab-webui/blob/83f580620f5e41efa9e9c1053574a3b904e8a052/src/containers/Sidebar/service.js#L19

3] Update of locales for ASAB-Config in demo app

ateska (approved):

PR #255, opened 2022-03-30
Possibility to pass a function to generate subpath in DataTable

petrKavulok:
Currently we generate links for this type of cells from objects : `link: { pathname: "my/pathname/", key: "_id" }`

And the result is  pathname + key  (e.g. `my/pathname/123` where `123` is `_id` )

This feature introduces another option to construct these paths with a function.n This will give us more freedom in future for generating those links and keep regular styles for those links without need to use customComponent

https://user-images.githubusercontent.com/79644454/160885443-bcecf358-863b-44cf-b704-1aafebcf556a.mp4


titkov49 (approved):

Pe5h4 (approved):
Ok, thats cool

PR #257, opened 2022-04-05
Create component for config validation by tenant

Pe5h4:
- Create global function, which validate configuration contra current tenant (if Tenant module is part of the application) and returns `true` if one of the tenant in the configuration match with current tenant or `false` if there is no match

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,8 @@
 
 - Add an option to construct a link in DataTable's cell using a function (INDIGO Sprint 220318, [!255](https://github.com/TeskaLabs/asab-webui/pull/255))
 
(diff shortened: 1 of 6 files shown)


Pe5h4:
@ateska In this ASAB-UI PR, a function which validate configuration with current tenant has been implemented
As per tuesday's workshop, I put this global function to `configuration` part of ASAB-UI. 
The function returns `true` if one of the tenant in the configuration match with current tenant or `false` if there is no match.
If tenant module is not part of the application, this function always returns `false` (so it means that the configuration should be visible to everyone)

I also implement that to tools, since its a part of ASAB-UI. In the gif image you can see the full implementation within BitSwan UI.

Documentation about how to implement this function is written directly in the .js file with the function and also in `config.md` docs.
https://github.com/TeskaLabs/asab-webui/blob/b6525a113880469d46d8ea645211003c38e78fc6/src/config/validateConfiguration.js#L3-L70

tenant-config-validation

ateska (approved):

PR #258, opened 2022-04-06
Make monaco loader part of the application

titkov49:
CHANGES: 
- `@monaco-editor/react` library was replaced with `react-monaco-editor`
- `monaco-editor-webpack-plugin` was added
- webpack `entries` and babel preset configuration were extended in order to enable editor workers
- `corejs@3` exact version was installed and specified (before was dependecy of other packages) in order to prevent versioning missmatch for babel


Pe5h4 (approved):

ateska (approved):

ateska:
Will this have an impact on the core application size?
Lazy loading is OK.

titkov49:
@ateska Hi,
No, it will not

PR #265, opened 2022-04-21
Move API call initialization from constructor of service to init part

Pe5h4:
- This is a bugfix, the initialization of Service API in the constructor part of the service was causing the not appending Authroization Header in the request. This initialization has been moved to init part of the service to prevent previously described issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -309,3 +309,5 @@
 - Add monaco loader plugin to webpack.build.js (INDIGO Sprint 220401, [!260](https://github.com/TeskaLabs/asab-webui/pull/260))
 
 - Fix library global styles (INDIGO Sprint 220401, [!262](https://github.com/TeskaLabs/asab-webui/pull/262))
+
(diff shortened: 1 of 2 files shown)


Pe5h4:
@ateska This is a fix on ASAB-UI sidebar service, which was missing Authorization Bearer token in the header of request to ASAB-Config service and thus it was causing problems, when `oauth2 introspect` was used in nginx configuration. The issue has been fixed by moving initialization of the ASAB API from constructor to the init part of the service.

asab-sidebar-auth-header-token

Here is the nginx settings for above gif

Selection_018

ateska (approved):
Is it respecting that the config is not called when asab-config is not present?

Pe5h4:
@ateska yes indeed https://github.com/TeskaLabs/asab-webui/blob/cd18ced4517e3bb321444a20b13ccf18a3db7ee0/src/containers/Sidebar/service.js#L18

PR #267, opened 2022-04-25
Refactor/credantials styles

Aringoaway:
add style in variable


src/containers/SplashScreen.js
```diff
@@ -2,9 +2,10 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 function SplashScreen(props) {
+	console.log(props, props.brand_image, "props");
```
petrKavulok, line 5: forgotten console.log

PR #268, opened 2022-04-26
Refactor/config authorization

Pe5h4:
- Refactor handling of `Authorization` in validateConfiguration function. The `Authorization` has been added to handle configuration with this section name. `:authorization` has been left as a fallback with TODO comment to remove it by the end of year 2022
- Refactor docu for Tools module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -246,6 +246,8 @@
 
 - Rename Library import and export endpoints (INDIGO Sprint 220419, [!266](http://gitlab.teskalabs.int/bitswan/bitswan-webui/-/merge_requests/266))
 
+- Refactor validateConfiguration function to be able to handle "Authorization" section name of configuration (INDIGO Sprint 220419, [!268](http://gitlab.teskalabs.int/bitswan/bitswan-webui/-/merge_requests/268))
+
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


Pe5h4:
@ateska In this ASAB-UI PR, refactorization of handling configurations with `Authorization` has been done.
As per discussion on review, configurations with authorization by tenant should have change their section names from `SectionName:authorization` to `Authorization` that it is unified across configurations. The option with `SectionName:authorization` has been left as a fallback (just in case) with TODO comment about removal at the end of 2022.

You can observve the change here: https://github.com/TeskaLabs/asab-webui/blob/36167406407cfea3e987a546e61083036e264dee/src/config/validateConfiguration.js#L77

This change should not have any impact on authorization functionality (and thus deployments), but we will have to gradually update our configurations and schemas in following redeployments of our already deployed products, where authorization is present.

After this change will be merged, BS-UI documentation and schemas will have to be updated (authorization is used in Discover and Dashboards) and updated also on docs.teskalabs.com

Selection_021

Another change in this PR is update of documentation for Tools module, where capital letters for `Tool` section name should be present in the first character (now it is lowecase).

Selection_022

Pe5h4:
@ateska

ateska (approved):

PR #283, opened 2022-06-02
Implement ControlledSwitchWithAuthz and UncontrolledSwitchWithAuthz

Pe5h4:
The purpose for this action is to have a switch component similar to `ButtonWithAuth`. This switch with authz can be used e.g. in Library module, where we want to avoid random people disabling and enabling configuration with the switcher. So with the usage of the switch with authz component, we can ensure, that only people with particular resources can provide operations with switch button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -269,6 +269,9 @@
 
 - MicroserviceDetailContainer Fallbacks and Container Info Update (INDIGO Sprint 220527, [!281](https://github.com/TeskaLabs/asab-webui/pull/281))
 
+- Delete Remove button from ASAB Config editor, add resource for config actions (INDIGO Sprint 220527, [!284](https://github.com/TeskaLabs/asab-webui/pull/284))
+
+
 ### Bugfixes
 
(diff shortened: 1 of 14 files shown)


Pe5h4:
@ateska This ASAB-UI PR extends usage of Switch components, that can be used similar way as ButtonWithAuthz.

2 new components `ControlledSwitchWithAuthz` and `UncontrolledSwitchWithAuthz` has been added to `auth` module. Those components use native `asab-webui` `ControlledSwitch` and `UncontrolledSwitch` components created in previous iterations.

The purpose of extending the switchers is the same as for `ButtonWithAuthz` - we want to prevent users without particular resources to trigger some action. E.g. Switch is used in Library and in there, we dont want users without particular rights to modify the files.

There is MR for ASAB Library in our gitlab, where `ControlledSwitchWithAuthz` has been implemented and it depends on this PR.

In this PR, unification of a code responsible for disabling the buttons has been done. So all authz components (`ButtonWithAuthz`, `ControlledSwitchWithAuthz` and `UncontrolledSwitchWithAuthz`) use the same process of disabling the buttons.

The documentation has been extended as well.

Here is the flow on the Library example:

implement-buttons-with-authz

ateska (approved):

PR #292, opened 2022-06-28
Refactor tenant selection card to align it with new design

Pe5h4:
Update Tenant selection card according to new design


Pe5h4:
tenantselectioncard

PR #298, opened 2022-06-29
collapsable header fix

petrKavulok:
https://user-images.githubusercontent.com/79644454/176433441-524d856c-af2a-4412-9500-c3cdbcc919f3.mov

### DIFF
--- a/src/containers/Header/index.js
+++ b/src/containers/Header/index.js
@@ -65,14 +65,14 @@ export function Header(props) {
 
 			}
 
-				{/* toggler */}
-				<div className={`header-props-toggler mt-1 mr-3 p-0 ${headerProperties && 'header-props-open' }`} onClick={() => setHeaderProperties(!headerProperties)}>
-					<i className="cil-chevron-bottom"></i>
-				</div>
+			{/* toggler */}
+			<div className={`header-props-toggler mr-3 p-0 ${headerProperties ? 'header-props-open' : '' }`} onClick={() => setHeaderProperties(!headerProperties)}>
+				<i className="cil-chevron-bottom"></i>
+			</div>
 			
 			{/* smallscreen menu  */}
 			{ headerProperties && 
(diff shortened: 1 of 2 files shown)


src/styles/components/header.scss
```diff
@@ -47,12 +57,31 @@
...
+		flex-direction: column;
+		height: 103px;
+		display: flex;
+		flex-direction: column;
+		justify-content: space-evenly;
+		transition: height 300ms ease;
+	}
+		
 	.header-props-toggler{
+		color: $link-color;
```
Pe5h4, line 80: @petrKavulok Toggler should be `$toggle-text-color`

src/containers/Header/index.js
```diff
@@ -65,14 +65,14 @@ export function Header(props) {
 
 			}
 
-				{/* toggler */}
-				<div className={`header-props-toggler mt-1 mr-3 p-0 ${headerProperties && 'header-props-open' }`} onClick={() => setHeaderProperties(!headerProperties)}>
-					<i className="cil-chevron-bottom"></i>
-				</div>
+			{/* toggler */}
+			<div className={`header-props-toggler mr-3 p-0 ${headerProperties ? 'header-props-open' : '' }`} onClick={() => setHeaderProperties(!headerProperties)}>
```
Pe5h4, line 69: @petrKavulok what was wrong with this className condition?

```
headerProperties && 'header-props-open'
```

Pe5h4 (changes requested):
Is there any way that we can move the application image to the middle when on small screens as it was before? If so, can you please create ticket for that and put it to backlog?

src/containers/Header/index.js
petrKavulok: it would for some reason return `class="header-props-toggler mr-3 p-0 false"`

Snímek obrazovky 2022-06-29 v 20 57 36

PR #323, opened 2022-07-13
fix padding/margin in Header nav

Aringoaway:
Header - Add bit more padding between the items on the right side of the header


Aringoaway:
before:
image

after:
image

PR #328, opened 2022-07-18
Data table content loader

Aringoaway:
fix color for content loader

### DIFF
--- a/src/components/DataTable/index.js
+++ b/src/components/DataTable/index.js
@@ -103,7 +103,7 @@ export function DataTable ({
 							rowClassName={customRowClassName}
 						/>}
 
-						{isLoading && contentLoader && <CellContentLoader cols={headers.length} rows={limit ?? 5} foregroundColor={theme === "dark" ? '#788dcb' : '#fff'} /> }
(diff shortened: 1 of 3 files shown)


Aringoaway:
@Pe5h4 can you review pls
datatable-loader

src/components/DataTable/index.js
```diff
@@ -103,7 +103,7 @@ export function DataTable ({
 							rowClassName={customRowClassName}
 						/>}
 
-						{isLoading && contentLoader && <CellContentLoader cols={headers.length} rows={limit ?? 5} /> }
+						{isLoading && contentLoader && <CellContentLoader cols={headers.length} rows={limit ?? 5} foregroundColor={theme === "dark" ? '#788dcb' : '#fff'} /> }
```
Pe5h4, line 106: @Aringoaway Is this change necessary? From the code I see if you dont specify foregroundColor, then the content loader will have the default and the default in content loader is the same as here

Pe5h4 (changes requested):
Also, please add this to changelog

src/components/DataTable/index.js
Aringoaway: @Pe5h4 fixed it

PR #329, opened 2022-07-18
class for margin top in cards

petrKavulok:
solution for all card's margin in seacat auth ui


Pe5h4:
@petrKavulok Whats up with this PR? Haven't we agreed on closing it?

PR #330, opened 2022-07-19
Microservices implement search, filter and sort by

Aringoaway:
Microservices - Implement search, pagination and sort by name into the list of Microservices


Pe5h4 (approved):
@Aringoaway It looks good, please add changelog, then it is mergeable

PR #332, opened 2022-07-20
Refactor/version and repo info

petrKavulok:
UserInterfaceCard shows current tag version and repository name

- local build
Snímek obrazovky 2022-07-20 v 16 49 50

- version from tag
Snímek obrazovky 2022-07-20 v 17 36 12


webpack/common.js
```diff
@@ -96,11 +96,23 @@ exports.convertKeysForHtml = function(obj){
...
+
+exports.getRepository = function(){
+	try {
+		const url = execSync("git config --get remote.origin.url", { encoding: 'utf8' }).toString();
+		const suburl = url.substring(url.lastIndexOf("/") + 1, url.lastIndexOf(".git"));
+		return suburl;
+	} catch (e) {
+		console.log("Error when getting repository from git. This error doesn't affect build process but you won't be able to see repository name");
+		console.error(e);
+		return "";
```
Pe5h4, line 116: @petrKavulok You can return `undefined` or you dont need to return anything at all, then it will be `undefined` by default if nothing is returned. Its because the condition here:

```
{ repository &&
		<React.Fragment>
```

In the code you wrote will not be rendered if `repository` is undefined, but it will be rendered, when repository will be `""` an empty string.

Pe5h4 (changes requested):
Good job! It looks awesome, just minor comment (please test that) and add changelog.

Pe5h4 (approved):

PR #333, opened 2022-07-21
Bugfix Configuration useForm

Aringoaway:
fix bug in Configuration in useForm, overwriting values in input


Aringoaway:
@Pe5h4 can you review pls

Pe5h4 (approved):
@Aringoaway Good. Have you tested that e.g. by saving default value, etc?

Aringoaway:
@Pe5h4 Yes, I tested this by adding, saving, removing data

PR #336, opened 2022-07-26
Move knowledge base component to a module

Pe5h4:
- modularize knowledge base componet


Pe5h4:
@ateska This is an attempt to modularize `Knowledge base` component of ASAB-UI, which in `main` branch of ASAB UI is a component and thus it demands `react-markdown` and `rehype-raw` libraries by default for every application (e.g. for SA UI and by that it makes the applications larger by size than it could be).

So after this refactorization, the Knowledge base is a module and we shook off `react-markdown` and `rehype-raw` libraryies, which saved us over 300kb of space (if we remove those libraries from package.json).

I think it is possible to diminish the size even more by upgrading webpack (this we are going to plan for a next sprint)

Here on example of SA UI:

Before:
Selection_066

After:
Selection_067

Here is how it look like when Knowledge base is a part of Logmanio:

knowledge-base-1

Knowledge base content is stored in `public` folder as it has been designed before. And I think implementing of Knowledge base as a module is now even easier than implementing it as a component (as it was before).

ateska (approved):
Very nice 👍🏻

PR #337, opened 2022-07-27
Refactor/Datatable column width

petrKavulok:
In this PR 
- default width of a column has been changed to `5em`
- `customStyle` prop has been renamed to `customCellStyle` as it only affects the data cells, on the table's header
- `customHeaderStyle` prop has been added to headers. It changes styles of the table header and can be used to change `min-width` and other css properties of the whole column.


petrKavulok:
Default settings without data
Snímek obrazovky 2022-07-27 v 22 37 25

Header with `customHeaderStyle: { minWidth: '100px' }` without data
Snímek obrazovky 2022-07-27 v 22 36 37

Header with `customHeaderStyle: { minWidth: '100px' }` including data
Snímek obrazovky 2022-07-27 v 22 32 34

src/components/DataTable/Table.js
```diff
@@ -126,7 +127,7 @@ const Headers = ({ headers, advmode, sublists }) => (
 			{headers.map((_, idx) =>
 				<col
 					className={`data-table-col${idx}`}
-					style={{ width: (idx === headers.length - 1) ? "auto" : "1px" }}
+					style={{ width: (idx === headers.length - 1) ? "auto" : "5em", ..._.customHeaderStyle }}
```
Pe5h4, line 130: @petrKavulok

1] shouldnt be `..._.customHeaderStyle` `...customHeaderStyle` ?

2] Do we want to preserve the condition on `width: "auto"` ? What is the benefit of it, if default should be `5em` as you've described in the docu?

Pe5h4:
> Default settings without data Snímek obrazovky 2022-07-27 v 22 37 25
> 
> Header with `customHeaderStyle: { minWidth: '100px' }` without data Snímek obrazovky 2022-07-27 v 22 36 37
> 
> Header with `customHeaderStyle: { minWidth: '100px' }` including data Snímek obrazovky 2022-07-27 v 22 32 34

@petrKavulok Why are those columns in Default settings without data squeezed as it used to be? Shouldn't there be a default width of the column set to `5em`? I guess it should be but from the image it is not. I think it has something to do with the condition on `width: auto` as I pointed out in previous post

src/components/DataTable/Table.js
petrKavulok: a summary of what was discussed on stand up
1.) we're maping headers on ln127. -> the underscore `_` represents each element (not sure why it is like that, I didn't change it since i thought it was some sort of legacy code :D )

2.) `width: auto` is assigned to the last column of the table to always fill the rest of the table

Pe5h4 (approved):
@petrKavulok Good, if tested, then lets merge it

PR #350, opened 2022-08-24
Fix of the Safari bug from the workshop.

ateska:
Fixed frozen refresh bug observed in Safari.

When `window.location.replace(...)` is issued on Safari, it can end up if frozen state.
The solution is to add `await new Promise(r => setTimeout(r, 3600 * 1000));` aka long running Promise just after that `replace` to ensure that Safari has enough time to initiate reload properly.

TODOs:

- [x] consider providing own function to do `window.location.replace + long running Promise`
- [x] `set_tenants` is now asynchronous, check if it is called correctly everywhere
- [x] `simulateUserinfo()` is now asynchronous, check if it is called this way everywhere


ateska:
Looks good from my end, feel free to approve & merge.

Pe5h4 (approved):

PR #356, opened 2022-09-05
DataTable dynamic item display

Aringoaway:
Dynamically set up items per page by the height of the screen
 
These changes add the ability to dynamically display the number of rows in the table on the first load. To do this, the height and setLimit must be passed to the table props. If no setLimit is added, no information will be displayed in the table. The table will be empty.

### DIFF
--- a/doc/datatable/datatable.md
+++ b/doc/datatable/datatable.md
@@ -705,10 +705,10 @@ The number of items displayed per page will be a multiple of 5.
 ```js
 import React, { useState, useEffect, useRef } from 'react';
 
-...
+	...
 
 function (props) {
-...
+	...
 	const [limit, setLimit] = useState(0);
 	const [height, setHeight] = useState(0);
 	const ref = useRef(null);
@@ -736,7 +736,7 @@ function (props) {
 			>
 		</div>
 	);
-..
+	..
 }
 ```
 
--- a/doc/datatable/datatable.md
+++ b/doc/datatable/datatable.md
@@ -290,7 +290,7 @@ Example of fetched data:
 
 # Optional
 
-`DataTable` can also accept optional props `limit`, `setLimit`, `createButton`, `buttonWithAuthz`, `customButton`, `customComponent`, `search`, `onSearch`, `isLoading`, `noItemsComponent`, `customCardBodyComponent`, `sublistsKey` and `onDownload`.
+`DataTable` can also accept optional props `limit`, `setLimit`, `createButton`, `buttonWithAuthz`, `customButton`, `customComponent`, `search`, `onSearch`, `isLoading`, `noItemsComponent`, `customCardBodyComponent`, `sublistsKey`, `onDownload`, `heightRef` and `setHeightRef`.
 
 Example of `DataTable` with all props:
 
@@ -313,6 +313,7 @@ Example of `DataTable` with all props:
 	onDownload={onDownload}
 	isLoading={isLoading}
 	noItemsComponent={noItemsComponent}
+	heightRef={heightRef}
 />
 ```
 
@@ -689,6 +690,37 @@ Property `onDownload` of `customComponent` is needed if you want to use custom c
 		generate: (row, header) => <SomeFancyComponent>{header.name} | {row.some_key}</SomeFancyComponent>,
 		onDownload: (row, header) => `${header.name} | ${row.some_key}`;
 	}
+}
+```
+
+Property `heightRef` and `setHeightRef` is needed if you want the item of elements in the table to adjust to the height of the web page when it first loads.
+NOTE: `heightRef` should be used in conjunction with the `limit` and `setLimit`.
+You will need to `useEffect`. This is where you will count the height of the container when you first load it. You will also need to `useRef`. You will need to place it on the wrapper tag. Be sure to style this tag to 100% height.
+The number of items displayed per page will be a multiple of 5.
+
+```js
+import React, { useState, useEffect, useRef } from 'react';
+
+...
+
+function (props) {
+	...
+	const [heightRef, setHeightRef] = useState(0);
+	const ref = useRef(null);
+	...
+	useEffect(() => {
+		setHeightRef(ref.current.clientHeight);
+	}, []);
+	...
+	return (
+		<div className="h-100" ref={ref}>
+			<DataTable
+				...
+				heightRef={heightRef}
+				...
+			>
+		</div>
+	);
 }
 ```
 
--- a/doc/datatable/datatable.md
+++ b/doc/datatable/datatable.md
@@ -695,9 +695,13 @@ Property `onDownload` of `customComponent` is needed if you want to use custom c
 
 Property `height` is needed if you want the item of elements in the table to adjust to the height of the web page when it first loads.
 NOTE: `height` should be used in conjunction with the `limit` and `setLimit`.
-You will need to `useEffect`. This is where you will count the height of the container when you first load it. You will also need to `useRef`. You will need to place it on the wrapper tag. Be sure to style this tag to 100% height.
+> You will need two `useEffect`:
+> 
+> - The first will calculate the height of the container when it is first loaded. 
+> - The second you need to have the number of rows displayed dynamically, you need to add a condition (`limit !== undefined`), where you call the function that gets the data for the table.
+> 
+You also need to add `useRef`. You will need to place it on the wrapper tag. Be sure to style this tag to 100% height.
 The number of items displayed per page will be a multiple of 5.
-
 ```js
 import React, { useState, useEffect, useRef } from 'react';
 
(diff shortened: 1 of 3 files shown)


Aringoaway:
DINHEI

Aringoaway:
@Pe5h4 can you review pls?

doc/datatable/datatable.md
```diff
@@ -290,7 +290,7 @@ Example of fetched data:
 
 # Optional
 
-`DataTable` can also accept optional props `limit`, `setLimit`, `createButton`, `buttonWithAuthz`, `customButton`, `customComponent`, `search`, `onSearch`, `isLoading`, `noItemsComponent`, `customCardBodyComponent`, `sublistsKey` and `onDownload`.
+`DataTable` can also accept optional props `limit`, `setLimit`, `createButton`, `buttonWithAuthz`, `customButton`, `customComponent`, `search`, `onSearch`, `isLoading`, `noItemsComponent`, `customCardBodyComponent`, `sublistsKey`, `onDownload`, `heightRef` and `setHeightRef`.
```
Pe5h4, line 293: @Aringoaway Where datatable accept `setHeightRef`? Since I dont see it anywhere :)

doc/datatable/datatable.md
```diff
@@ -692,6 +693,37 @@ Property `onDownload` of `customComponent` is needed if you want to use custom c
...
+	...
+	useEffect(() => {
+		setHeightRef(ref.current.clientHeight);
+	}, []);
+	...
+	return (
+		<div className="h-100" ref={ref}>
+			<DataTable
+				...
+				heightRef={heightRef}
```
Pe5h4, line 719: @Aringoaway I suggest to rename `heightRef` to `height` e.g.

```
height={...}
```

src/components/DataTable/index.js
```diff
@@ -52,6 +59,14 @@ export function DataTable ({
 		}, 500);
 	}, [filterValue]);
 
+	// rounding page number divisible by 5
+	function round5(x) {
```
Pe5h4, line 63: @Aringoaway rename the function to `round` pls

src/components/DataTable/index.js
```diff
@@ -30,16 +30,23 @@ export function DataTable ({
...
 	const [filterValue, setFilterValue] = useState('');
 	const [isLimitOpen, setLimitDropdown] = useState(false);
 	const timeoutRef = useRef(null);
 	const [countDigit, setCountDigit] = useState(1)
 	
 	const { t } = useTranslation();
 
+	useEffect(() => {
+		if ((heightRef && setLimit) && heightRef !== 0) {
+			let tableRowCount = Math.floor((heightRef - 200)/48);
```
Pe5h4, line 45: @Aringoaway Please add comment what numbers `200` and `48` means

src/components/DataTable/index.js
```diff
@@ -30,16 +30,23 @@ export function DataTable ({
 	customButton, customComponent,
 	customRowStyle, customRowClassName,
 	customCardBodyComponent,
-	limitValues = [10, 15, 25, 50],
-	contentLoader = true, category
-	}) {
+	limitValues = [5, 10, 15, 20, 50],
```
Pe5h4, line 33: @Aringoaway What is the default? I mean when there is no height passed to the index.js, there should be some default,, right? I propose `10`

Pe5h4 (changes requested):

Aringoaway:
dyn

src/components/DataTable/index.js
```diff
@@ -52,10 +63,20 @@ export function DataTable ({
 		}, 500);
 	}, [filterValue]);
 
+	// rounding page number divisible by 5
+	function round(x) {
+		const rowCount = Math.round(x / 5) * 5;
```
Pe5h4, line 68: @Aringoaway To make it more defensive, you need to ensure that value `x` is an integer (not string, not boolean, not undefined, etc..). If not, return default limit

src/components/DataTable/index.js
```diff
@@ -52,10 +63,20 @@ export function DataTable ({
 		}, 500);
 	}, [filterValue]);
 
+	// rounding page number divisible by 5
+	function round(x) {
```
Pe5h4, line 67: @Aringoaway Pls rename this function to something more understandable. When you think of it, what does this function do? It certainly does not "round", but rather returns count of the rows in datatable

Pe5h4 (changes requested):

doc/datatable/datatable.md
```diff
@@ -692,6 +693,40 @@ Property `onDownload` of `customComponent` is needed if you want to use custom c
...
+
+...
+
+function (props) {
+	...
+	const [limit, setLimit] = useState(10);
+	const [height, setHeight] = useState(0);
+	const ref = useRef(null);
+	...
+	useEffect(() => {
```
Pe5h4, line 712: @Aringoaway I advice you to extend the description here as we spoke of on our workshop :)

Aringoaway:
table

Aringoaway:
This is what will happen if there is no height or setLimit in the props
condition

src/components/DataTable/index.js
```diff
@@ -52,10 +63,25 @@ export function DataTable ({
...
+	function roundedNumRows(x) {
+		if (Number.isInteger(x)) {
+			const rowCount = Math.round(x / 5) * 5;
+			if (rowCount <= 0) {
+				return 5;
+			} else {
+				return rowCount;
+			}
+		} else {
+			return limit;
```
Pe5h4, line 76: @Aringoaway Are you sure about this? Would not it be overwritten by `limit` props passed to `DataTable` component as you have it in the example in the documentation? And since in example you pass `undefined`, it can happen, that this will return `undefined` as well... and I dont think this is desired

Pe5h4 (changes requested):

src/components/DataTable/index.js
Pe5h4: @Aringoaway

https://github.com/TeskaLabs/asab-webui/pull/356#issuecomment-1245032113

This is happening exactly because of that you are returning `undefined` as I described in my comment in this thread

Pe5h4: You should return `default` value here and the default is 10 if I am not mistaken...

Pe5h4:
@Aringoaway There is however another way of doing that:

```
function (props) {
	...
	const [limit, setLimit] = useState(0);
	const [height, setHeight] = useState(0);
	const ref = useRef(null);
	
	useEffect(() => {
		setHeight(ref.current.clientHeight);
	}, []);

	useEffect(() => {
		if (limit > 0) {
			fetchData();
		}
	}, [limit]);

        const fetchData = () => {...}

	return (
		<div className="h-100" ref={ref}>
			<DataTable
				...
				limit={limit}
				setLimit={setLimit}
				height={height}
				...
			>
		</div>
	);

..
```

doc/datatable/datatable.md
```diff
@@ -692,6 +693,53 @@ Property `onDownload` of `customComponent` is needed if you want to use custom c
...
+			fetchData();
+		}
+	}, [limit]);
+
+	const fetchData = () => {...}
+
+	return (
+		<div className="h-100" ref={ref}>
+			<DataTable
+			...
```
Pe5h4, line 731: @Aringoaway Indentation here pls

src/components/DataTable/index.js
```diff
@@ -52,10 +63,25 @@ export function DataTable ({
...
+	function roundedNumRows(x) {
+		if (isNaN(x) == false) {
+			const rowCount = Math.round(x / 5) * 5;
+			if (rowCount <= 0) {
+				return 5;
+			} else {
+				return rowCount;
+			}
+		} else {
+			return 10;
```
Pe5h4, line 76: @Aringoaway Add a comment here that it is a default limit value

src/components/DataTable/index.js
```diff
@@ -52,10 +63,25 @@ export function DataTable ({
 		}, 500);
 	}, [filterValue]);
 
+	// rounding page number divisible by 5
+	function roundedNumRows(x) {
+		if (isNaN(x) == false) {
+			const rowCount = Math.round(x / 5) * 5;
+			if (rowCount <= 0) {
+				return 5;
```
Pe5h4, line 71: @Aringoaway Add a comment here that it is a lowest limit value

Pe5h4 (changes requested):
@Aringoaway Alright, it looks cool! Only minor stuff to refactor. Has it been also tested when navigating to the screen with datatable from different part of the application?

src/components/DataTable/index.js
Aringoaway: > @Aringoaway Alright, it looks cool! Only minor stuff to refactor. Has it been also tested when navigating to the screen with datatable from different part of the application?

@Pe5h4 I have tested this on other tables and everything works

Pe5h4 (approved):
@Aringoaway It looks cool! Does content loader work when using dynamic table option?

Aringoaway:
> @Aringoaway It looks cool! Does content loader work when using dynamic table option?

@Pe5h4 Yes, content loader work

PR #358, opened 2022-09-14
Refactor form-text color

Aringoaway:
I have replace this style to make it suitable for all the examples where it will be used.
This style uses the <FormText> components.
This change comes with a marge-request in the SeaCat WebUI.


Pe5h4 (commented):
@Aringoaway How this change will affect other applications using `form-text`

Aringoaway:
@Pe5h4 everything will remain as it is, it will just work in all cases where this component is used

Pe5h4 (approved):

PR #360, opened 2022-09-19
Fix button link color

Aringoaway:
Fix button link color when state active or focus


Aringoaway:
@Pe5h4 can you review it pls? 
btnfix

Aringoaway:
@Pe5h4 I checked the colours and they are all the same in the light and dark theme
you can see it in this video in the tooltip above the button
colors

Pe5h4 (approved):

PR #361, opened 2022-09-20
Fix DataTable Docs

Aringoaway:
Fix italic style in Option, add ` for  tags

I received feedback from Peter and I will correctthe dynamic display of the number of items in DataTable description to make it clearer


Aringoaway:
I fixed sprint number for MR #358 because I merged it yesterday

PR #364, opened 2022-09-23
Bugfix/login without tenant

Aringoaway:
I tested it:
- without AuthModule
- without SeaCatAuthModule a + tenant [ ] and tenant [firsttenant, secondtenant]
- without devConfig
- without tenant in devConfig

### DIFF
--- a/src/modules/tenant/service.js
+++ b/src/modules/tenant/service.js
@@ -40,6 +40,11 @@ export default class TenantService extends Service {
 
 		// In case if the tenant list from userinfo is undefined or empty remove tenant parameter from URL
 		if ((!tenants_list) || (tenants_list.length == 0)) {
+			this.App.Store.dispatch({
+				type: types.TENANTS_CHANGED,
+				undefined,
+				current: null
+			});
 			await locationReplace(window.location.pathname + window.location.hash);
 			return;
 		}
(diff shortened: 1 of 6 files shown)


Aringoaway:
If there is no user without a tenant, they will be shown a card with an 'Back to login' button.
https://user-images.githubusercontent.com/65866093/194518485-3e8f2a14-a262-48d2-aeea-e39963338248.mp4

src/modules/tenant/service.js
```diff
@@ -40,6 +40,11 @@ export default class TenantService extends Service {
 
 		// In case if the tenant list from userinfo is undefined or empty remove tenant parameter from URL
 		if ((!tenants_list) || (tenants_list.length == 0)) {
+			this.App.Store.dispatch({
+				type: types.TENANTS_CHANGED,
+				undefined,
+				current: null
+			});
```
ateska, line 47: Is this needed?

ateska (changes requested):
Small remark.

src/modules/tenant/service.js
Aringoaway: @ateska Yes, is an important part of this change

Aringoaway: @ateska we need it at a time when no tenants are assigned. And we write it as tenant - undefined to distinguish situations when to show this Logout card

ateska: I still don't get it.
Let's discuss this latest on review, pls.

ateska (changes requested):

Pe5h4:
This branch is staled

ateska:
Yeah. Close it if it is not needed to be merged.

I still don't get why we need to set a tenant and then refresh the application ... that was my only serious question here.

Pe5h4:
@Aringoaway ^^^

PR #366, opened 2022-09-27
Refactor logout with openidconnect

Pe5h4:
- Refactor logout from tenant selection card with use of `/openidconnect/logout` endpoint


Pe5h4:
@ateska This is a refactorisation of TenantSelection card logout screen with correct logout behavior. The `/public/logout` request was replaced with `/openidconnect/logout` taken from Auth module, where it is already implemented.

The action on `Back to login` button check, if AuthModule is present within the application and only then triggers the action.

logout_oidc

ateska (approved):

PR #367, opened 2022-09-29
Bugfix/spinner style

Aringoaway:
when class will be added, spinner-off class turn it off


Aringoaway:
@Pe5h4 
off-p

Pe5h4 (approved):
@Aringoaway Ok, it looks better. Can we move the spinner a bit to the center of the page (in the next iteration)? Be careful with that since on mobile screen, we dont want to have it somewhere in the bottom. I think you would need to use media screen size for that

PR #368, opened 2022-10-04
Access Control adv mode

Aringoaway:
add advanced mode to show JSON

### DIFF
--- a/src/modules/auth/AccessControlScreen.js
+++ b/src/modules/auth/AccessControlScreen.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { connect } from 'react-redux';
+import {connect, useSelector} from 'react-redux';
+import ReactJson from "react-json-view";
 
 import {
 	Container, Row, Col,
-	Card, CardHeader, CardTitle, CardSubtitle, CardBody, CardFooter,
-	Button
+	Card, CardHeader, CardTitle, CardSubtitle, CardBody
 } from 'reactstrap';
 
 import "./styles.scss";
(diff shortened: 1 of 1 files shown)


Aringoaway:
image
adv

src/modules/auth/AccessControlScreen.js
```diff
@@ -54,51 +55,84 @@ function AccessControlCard(props) {
...
+								<div className="text-center">
+									{t("AccessControlScreen|The user information is invalid, you session is likely expired.")}
+								</div>
+							)}
+						</CardBody>
+					</Card>
+				</Col>
+			</Row>
+
+			<Row className="justify-content-md-center mt-4">
```
Pe5h4, line 110: @Aringoaway I think it would be better to wrap the whole `Row` to `advmode` than just the content of the row. Also I am not sure about the justification of the content of the Row (pls take a look how other advanced mode cards look like in other screens)

Pe5h4 (changes requested):

Aringoaway:
@Pe5h4 I redid it :)
image

src/modules/auth/AccessControlScreen.js
```diff
@@ -54,51 +55,75 @@ function AccessControlCard(props) {
...
+						</>
+					) : (
+						<div className="text-center">
+							{t("AccessControlScreen|The user information is invalid, you session is likely expired.")}
+						</div>
+					)}
+				</CardBody>
+			</Card>
+			{advmode &&
+				<Card className="mt-4">
```
Pe5h4, line 106: @Aringoaway add `className="mt-4 shadow animated fadeIn"` that the card is similar to the one above - when you look at it, the adv card does not have a shadow, etc. so it looks kinda strange in comparison with the card abut access control

Pe5h4 (changes requested):

Pe5h4 (approved):

PR #370, opened 2022-10-06
Implement tenant sessions

Pe5h4:
Related to: https://github.com/TeskaLabs/seacat-auth/pull/92


Pe5h4:
@ateska This PR is an implementation of tenant sessions to ASAB WebUI as per changes in SeaCat Auth service: https://github.com/TeskaLabs/seacat-auth/pull/92

- the access denied card has been implemented
- access is denied for users without particular tenant
- when access is denied, user can navigate to login (with base url - so no redirect)
- when user has no tenant at all, they cannot enter the application at all (when tenant module is present)
- so this PR also solve the issue closed in this PR https://github.com/TeskaLabs/asab-webui/pull/364 (aka login without tenant)
- this PR is related to https://github.com/TeskaLabs/seacat-auth/pull/92 and its a breaking change, so we need to align versions and make a breaking change release note (when released)

flow as a superuser:
superuser

flow as administrator which is trying to access tenant which they dont have assigned
seacat-admin

flow as user without any tenant trying to access the application with tenant which they dont have assigned and later from position when they have been navigated back to login (so no tenant specified in the url)
no-tenant-user

flow when switched to nonexisten tenant
change-to-nonexistent-tenant

ateska:
Looks cool.

The "Access denied" card needs some spacing ...

ateska (approved):

PR #373, opened 2022-11-01
CardHeader button z-index

Aringoaway:
Before:
the button popped out on CardBody
image

After:

image

### DIFF
--- a/src/styles/components/button.scss
+++ b/src/styles/components/button.scss
@@ -476,6 +476,7 @@ $btn-dark-outline-hover-bg-color: var(--btn-dark-outline-hover-bg-color);
 .btn-group {
 	&>.btn {
 		flex: none;
+		z-index: 0 !important;
 	}
 	&>.btn:not(:first-child) {
 		margin-left: 0;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Release Candidate
 
+### Bugfixes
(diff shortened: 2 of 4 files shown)


Pe5h4 (commented):
@Aringoaway Shouldnt be here rather some card header bottom padding (or flexbox) for those cases, instead of just hiding the part of the button?

Aringoaway:
This MR goes with this [MR asab-library-webui ](http://gitlab.teskalabs.int/asab/asab-library-webui/-/merge_requests/31)

Pe5h4 (approved):

src/styles/components/button.scss
```diff
@@ -476,6 +476,7 @@ $btn-dark-outline-hover-bg-color: var(--btn-dark-outline-hover-bg-color);
 .btn-group {
 	&>.btn {
 		flex: none;
+		z-index: 0 !important;
```
Pe5h4, line 479: @Aringoaway Are you sure about `z-index: 0` ?

Pe5h4 (changes requested):

src/styles/components/button.scss
Aringoaway: I have now tested these changes and decided to remove this branch completely, because removing the `height` in [this request](http://gitlab.teskalabs.int/asab/asab-library-webui/-/merge_requests/31) completely solves the problem

PR #377, opened 2022-11-22
Fix Card-Header styles

Aringoaway:
Fix broken widget card-header

Before:
image

After:
image

After changing the styles, the headers look exactly the same and their functionality is intact. I have tested this with all the variants we have in the app

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Release Candidate
 
+### Bugfixes
+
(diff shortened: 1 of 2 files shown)


Aringoaway:
This MR comes together with this [MR](http://gitlab.teskalabs.int/bitswan/bitswan-webui/-/merge_requests/183)

Pe5h4 (changes requested):
@Aringoaway Please add changelog

Pe5h4 (approved):

Pe5h4:
PLease resolve conflicts in changelog

PR #382, opened 2023-01-03
Refactor bredcrumbs

Aringoaway:
### Before:
breadcrumbs_issue-exact-match

### After:
https://user-images.githubusercontent.com/65866093/210371647-2a81fbf1-8056-49eb-b23a-cfb7b4e0d14f.mp4

### DIFF
--- a/src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
+++ b/src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
@@ -9,14 +9,21 @@ const Breadcrumbs = ({
 	routes, match, app, disableContainerBreadcrumbs
 }) => {
 	const { t } = useTranslation();
-	// Get new crumbs each time location has been changed
-	const crumbs = routes.filter(({ path }) => match.path.includes(path)).
+	const crumbs = routes.filter(({ path }) => {
+			if ((path.split("/").length - 1) === (match.path.split("/").length - 1)) { // Checking if the number of slash ('/') is the same
+				if (match.path.includes(path) && (path.match(/[^\/]+$/)[0] === match.path.match(/[^\/]+$/)[0])) { // I take the string after the last slash ('/') and check if they match each other
+					return path;
+				}
+			} else {
+				return match.path.includes(path);
+			}
+		}).
 		map(({ path, ...rest }) => ({
 			path: Object.keys(match.params).length ?
 				Object.keys(match.params).reduce((path, param) =>
 					path.replace(`:${param}`, match.params[param]), path
 				) : path,
-				...rest
+			...rest
 		})).
 		filter(crumb => crumb.name);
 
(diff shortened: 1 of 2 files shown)


src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -26,13 +33,14 @@ const Breadcrumbs = ({
 	// TODO: Add disabling breadcrumbs
 	return (
 		<div className="breadcrumbs p-auto">
+			{console.log(crumbs, "crumbs")}
```
Pe5h4, line 36: @Aringoaway Remove console.log

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -9,14 +9,21 @@ const Breadcrumbs = ({
 	routes, match, app, disableContainerBreadcrumbs
 }) => {
 	const { t } = useTranslation();
-	// Get new crumbs each time location has been changed
-	const crumbs = routes.filter(({ path }) => match.path.includes(path)).
+	const crumbs = routes.filter(({ path }) => {
+			if ((path.split("/").length - 1) === (match.path.split("/").length - 1)) { // Checking if the number of slash ('/') is the same
```
Pe5h4, line 13: @Aringoaway Why cant you do just:

```
(path.split("/").length) === (match.path.split("/").length)
```
?

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -9,14 +9,21 @@ const Breadcrumbs = ({
...
+				if (match.path.includes(path) && (path.match(/[^\/]+$/)[0] === match.path.match(/[^\/]+$/)[0])) { // I take the string after the last slash ('/') and check if they match each other
+					return path;
+				}
```
Pe5h4, line 16: @Aringoaway Add also else condition, since if the condition is not met, it will not return anything
```
} else {
	return match.path.includes(path);
}
```
alternatively, you can do this:

```
if (match.path.includes(path) && (path.match(/[^\/]+$/)[0] === match.path.match(/[^\/]+$/)[0])) {
	return path;
}
return match.path.includes(path);
```

Pe5h4 (changes requested):
@Aringoaway It looks good, but you will need to test that properly (situations with no breadcrumbs, only one bredcrumb or some advanced breadcrumbs) and check the performance (this seems like a costly operation.

You may use `useMemo` for this operation, since it helps with the performance

```
const crumbs = useMemo(() => {
...
},[routes])
```

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -9,16 +9,25 @@ const Breadcrumbs = ({
 	routes, match, app, disableContainerBreadcrumbs
 }) => {
 	const { t } = useTranslation();
-	// Get new crumbs each time location has been changed
-	const crumbs = routes.filter(({ path }) => match.path.includes(path)).
+	const crumbs = useMemo(() => {
+		return routes.filter(({ path }) => {
+			if ((path.split("/").length === match.path.split("/").length) && match.path.includes(path)) { // Checking if the number of slash ('/') is the same and if it is included in the match.path
```
Pe5h4, line 14: @Aringoaway Add the comment above the `if` condition

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -9,16 +9,25 @@ const Breadcrumbs = ({
...
+				if ((path.match(/[^\/]+$/)[0] === match.path.match(/[^\/]+$/)[0])) { //  Take the string after the last slash ('/') and check if they match each other
```
Pe5h4, line 15: @Aringoaway Add the comment above the if condition (for better readability)

Pe5h4 (changes requested):
@Aringoaway It looks good! Minor stuff to change and also add a changelog

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
Pe5h4: Ok, thanks for explanation

Pe5h4 (approved):

PR #389, opened 2023-02-15
Switch sidebar on small and large screens

Aringoaway:
### Changes:
1. small screen > 768px sidebar sidebar is a modal window
2. 768 < medium screen <= 1024 is a collapsed sidebar
3. 1024 < large screen is a normal open sidebar
4. Reducing the margin when the sidebar is collapsed

https://user-images.githubusercontent.com/65866093/219601011-39b6de2d-6fca-4d1e-8f19-aa3e936cd9ee.mp4

### DIFF
--- a/src/containers/Sidebar/index.js
+++ b/src/containers/Sidebar/index.js
@@ -1,18 +1,23 @@
-import React, { useMemo, useEffect, useState } from 'react';
-import { connect, useSelector } from 'react-redux';
+import React, {useMemo, useEffect, useState} from 'react';
+import {connect, useDispatch, useSelector} from 'react-redux';
 
 import {Modal, Nav} from 'reactstrap';
 import SidebarItem from './SidebarItem';
 import NavbarBrand from './NavbarBrand';
 import SidebarBottomItem from './SidebarBottomItem';
+import {COLLAPSE_SIDEBAR} from "../../actions";
 
 
 const Sidebar = (props) => {
 	const [modal, setModal] = useState(false);
 	const isSidebarCollapsed = useSelector(state => state.sidebar.isSidebarCollapsed);
 	const [windowDimensions, setWindowDimensions] = useState({width: window.innerWidth});
+	const [bigScreen, setBigScreen] = useState(false);
+	const [mediumScreen, setMediumScreen] = useState(false);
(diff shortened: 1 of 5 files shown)


Aringoaway:
@Pe5h4 Can you review pls?:)
 I tested it in all the examples

src/containers/Sidebar/index.js
```diff
@@ -22,6 +27,49 @@ const Sidebar = (props) => {
 		return () => window.removeEventListener('resize', handleResize);
 	}, [windowDimensions]);
 
+	useEffect(() => {
```
Pe5h4, line 30: @Aringoaway Ok,everything from line 30 to line 71 can be simplified as:

```
useEffect(() => {
	if (windowDimensions.width && (windowDimensions.width > 768) && (windowDimensions.width <= 1024)) {
		dispatch({
			type: COLLAPSE_SIDEBAR,
			isSidebarCollapsed: true
		});
	} else {
		dispatch({
			type: COLLAPSE_SIDEBAR,
			isSidebarCollapsed: false
		});
	}
}, [windowDimensions.width])
```

Its that simple :) 

You are doing unnecessary dispatching based on state, which will have an impact on performance and thus it is not sustainable.

Pe5h4 (changes requested):

src/containers/Sidebar/index.js
Aringoaway: @Pe5h4 I understand, I also wrote the code this way at first. But I was confused by the behaviour of the sidebar. If I open it and start reducing the screen size, it keeps closing. Is this the correct behaviour shown in the video below? I've been trying to achieve the other behaviour shown in the video in the description
https://user-images.githubusercontent.com/65866093/219315266-afb3498f-d62a-4a61-9cab-d8991c1c25f5.mp4

Pe5h4: @Aringoaway  Yes, this should be the correct behaviour :)

Pe5h4 (changes requested):
@Aringoaway Update as per discussion here: https://github.com/TeskaLabs/asab-webui/pull/389/files#r1108160650

Pe5h4 (approved):
@Aringoaway Looks good from my point of view. If it has been tested on various browsers, then its mergable

PR #395, opened 2023-03-06
remove unused action

Aringoaway:
remove action `SET_SMALL_SIDEBAR` and `isSmallSidebarOpen`

### DIFF
--- a/src/containers/Sidebar/SidebarBottomItem.js
+++ b/src/containers/Sidebar/SidebarBottomItem.js
@@ -12,11 +12,8 @@ import { COLLAPSE_SIDEBAR } from "../../actions";
 
 import Icon from './SidebarIcon';
 
-import { SET_SMALL_SIDEBAR } from '../../actions';
-
 const SidebarBottomItem = ({ item, sidebarLogo }) => {
 	const isSidebarCollapsed = useSelector(state => state.sidebar.isSidebarCollapsed);
-	const isSmallSidebarOpen = useSelector(state => state.sidebar.isSmallSidebarOpen);
 
 	const location = useLocation();
 	const history = useHistory();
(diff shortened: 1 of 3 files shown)


src/containers/Sidebar/SidebarBottomItem.js
```diff
@@ -31,7 +28,6 @@ const SidebarBottomItem = ({ item, sidebarLogo }) => {
 		// or if item should always be uncollapsed
 		else if (item.children && !uncollapseAll) setOpen(prev => !prev);
 		// Close small sidebar on item click
```
Pe5h4, line 30: @Aringoaway You left the comment here, which is now unrelated. Remove it (also in the other file)

Pe5h4 (changes requested):

src/containers/Sidebar/SidebarBottomItem.js
```diff
@@ -27,11 +24,6 @@ const SidebarBottomItem = ({ item, sidebarLogo }) => {
 		// Preserve from history pushing when item.url doesn't exist
 		// or when current location pathname is the same as item.url
 		if (item.url && location.pathname !== item.url) history.push(item.url);
-		// Preserve from collapsing when item doesn't have children
-		// or if item should always be uncollapsed
-		else if (item.children && !uncollapseAll) setOpen(prev => !prev);
-		// Close small sidebar on item click
-		if (isSmallSidebarOpen && window.innerWidth < 768) dispatch({ type: SET_SMALL_SIDEBAR });
 	}
```
Aringoaway, line 27: I also removed this condition `else if (item.children && !uncollapseAll) setOpen(prev => !prev);`, because we do not pass these parameters as props

Pe5h4: @Aringoaway Alright, then remove it also from here, if its not used: https://github.com/TeskaLabs/asab-webui/blob/da1b44afc9ae2328aa05adae6190769b30e0dcc7/src/containers/Sidebar/SidebarItem.js#L47

Pe5h4 (changes requested):

Pe5h4 (approved):

PR #398, opened 2023-03-20
Update styles for arrow icon in the DataTable

Aringoaway:
### Changes

1. Add cursor pointer for icon
2. Add primary color for icon

https://user-images.githubusercontent.com/65866093/226354618-1cdfc877-266d-400c-95f5-dc8adba78171.mp4

### DIFF
--- a/src/components/DataTable/table.scss
+++ b/src/components/DataTable/table.scss
@@ -46,8 +46,11 @@ thead .data-table-tr th {
 
     & .data-table-tr {
         height: 48px;
-        & .data-table-th {
-            padding: auto;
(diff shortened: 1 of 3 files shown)


src/components/DataTable/table.scss
```diff
@@ -46,8 +46,11 @@ thead .data-table-tr th {
 
     & .data-table-tr {
         height: 48px;
-        & .data-table-th {
```
Aringoaway, line 49: I deleted this style because it's (auto) the wrong value for padding. This style doesn't work. I tested it, it works exactly the same without it

PR #405, opened 2023-04-03
Implement unauthorized access screen

Pe5h4:
- Implement Unauthorized access screen which will block users to enter the containers without particular resource

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 - Implement HelpComponent which shows the page description (INDIGO Sprint 230317, [!393](https://github.com/TeskaLabs/asab-webui/pull/393))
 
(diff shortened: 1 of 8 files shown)


Pe5h4:
@ateska In this PR, UnauthorizedAccess screen has been implemented. It will display information to unauthorized users (users without particular access rights given by the resources) that content of the page is not available without required resource.

This feature is applied on Routing of the containers, so it "blocks" the route(s), which require any access resource and when user is not authorized by a resource when entering some specific container, then UnauthorizedAccess screen is displayed. Example how to set up the feature is here: https://github.com/TeskaLabs/asab-webui/blob/c37a183abd6002dc5a4d9853918aa981f1b42f4e/doc/unauthorized-access.md

Maybe you will have some tips about the content displayed to the unauthorized users, since I am not sure if the proposed information is suitabe

```
You are not authorized to access this part of the application. Please ask your application administrator for resource
<re:sour:ce>
```

unauthorized-access

ateska (changes requested):
Text change, otherwise it looks cool

Pe5h4:
@ateska I have addded a colon and changed a text a bit:
Selection_135

PR #410, opened 2023-04-14
PRevent infinite userinfo loop when timeout time is zero or negative

Pe5h4:
### This fix should prevent infinite userinfo loops caused by check on expired session

- prevent "infinite" userinfo loop on session expiration, when the expiration time is set too small (1s or smaller)
- The fix is implementation of condition on `timeout <= 0` and triggering and alert message + setting up a timeout to 30s (due to session removals on backend once per minute)


Pe5h4:
@ateska Fix on "infinite" userinfo caused by negative timeout values

ateska (approved):

PR #414, opened 2023-04-27
Refactor/webpack-errors-removal-update

petrKavulok:


webpack/webpack.dev.js
```diff
@@ -127,6 +127,7 @@ module.exports = {
 						},
 					}
 				},
+				runtimeChunk: 'single',
```
Pe5h4, line 130: @petrKavulok why this? I thought `single` is a default value

Pe5h4 (changes requested):

webpack/webpack.dev.js
Pe5h4: @petrKavulok Ok, understand. Another question - why is it in webpack dev then?

Pe5h4 (changes requested):

webpack/webpack.dev.js
Pe5h4: @petrKavulok Alrighty

webpack/common.js
```diff
@@ -63,14 +63,16 @@ exports.getRules = function(config) {
 			test: /\.(png|jpe?g|gif|svg)(\?.*)?$/,
 			type: 'asset/resource',
 			generator: {
-				filename: 'assets/svg/[name][contenthash:8][ext]'
+				filename: 'svg/[name][contenthash:8][ext]',
+				outputPath: mode === 'dev' ? '' : 'assets/'
```
Pe5h4, line 67: @petrKavulok Not sure if this will not lead to a bigger confusion, as from my subjective point of view, I would keep `outputPath` separated and not mixing it in the common file. (But I understand your intention)

Pe5h4 (commented):
@petrKavulok Just a comment about keeping outputPath separated

Pe5h4 (approved):
This looks fine from my point of view (actually I thought that it is possible to have `test`, `type` and `generator: filename` in common, and `generator:outputPath` set separately for build and dev)

petrKavulok:
> This looks fine from my point of view (actually I thought that it is possible to have `test`, `type` and `generator: filename` in common, and `generator:outputPath` set separately for build and dev). PLease clarify with Ales

After discussion with Ales, we agreed to add a `TODO: ...` and tackle this issue during the MicroFrontends refactorization

PR #415, opened 2023-05-03
Refactor/lazy-loading

petrKavulok:
### in this PR

* lazy loading initiation in About, Auth,  Tools & Maintenance modules

### DIFF
--- a/src/modules/auth/index.js
+++ b/src/modules/auth/index.js
@@ -1,11 +1,15 @@
+import { lazy } from 'react';
 import Module from '../../abc/Module';
 
 import HeaderComponent from './header'
 import reducer from './reducer';
 import { types } from './actions'
 import { SeaCatAuthApi, GoogleOAuth2Api } from './api';
-import AccessControlScreen from './AccessControlScreen';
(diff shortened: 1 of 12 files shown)


src/modules/auth/index.js
```diff
@@ -1,11 +1,15 @@
...
 import reducer from './reducer';
 import { types } from './actions'
-import { SeaCatAuthApi, GoogleOAuth2Api } from './api';
-import AccessControlScreen from './AccessControlScreen';
-import { locationReplace } from 'asab-webui';
+import { SeaCatAuthApi } from './api';
+import { locationReplace, componentLoader } from 'asab-webui';
+const AccessControlScreen = lazy(() => componentLoader(() => import('./AccessControlScreen')));
+
+import "./components/unauthorizedaccess.scss";
```
Pe5h4, line 11: @petrKavulok This import does not have to be here, since unauthorized access is not loaded within auth index module

Pe5h4 (changes requested):
Minor change request, otherwise gut!

Pe5h4 (approved):
@petrKavulok Seems alright to me. If tested, then it can be merged

PR #421, opened 2023-05-18
Implement TitleService

Aringoaway:
### Changes
1. Implement TitleService
2. Update CHANGELOG

https://github.com/TeskaLabs/asab-webui/assets/65866093/e940a782-e3ac-489e-b008-58175c8264b9


Pe5h4 (approved):
@Aringoaway This looks very good

src/services/TitleService.js
```diff
@@ -0,0 +1,13 @@
+import Service from '../abc/Service';
+export default class TitleService extends Service {
+	constructor(app, serviceName="TitleService"){
+		super(app, serviceName);
+		this.App = app;
+		this.title = this.App.props.configdefaults?.title;
```
Pe5h4, line 6: @Aringoaway You dont want to take the title from `configdefaults` but from `config`. The best option is to use `this.App.Config.get("title")`

Pe5h4 (changes requested):

src/services/TitleService.js
```diff
@@ -0,0 +1,13 @@
+import Service from '../abc/Service';
+export default class TitleService extends Service {
+	constructor(app, serviceName="TitleService"){
+		super(app, serviceName);
+		this.App = app;
+	}
+
+	setTitle = (subTitle) => {
+		let title = this.App.Config.get("title");
```
Aringoaway, line 9: @Pe5h4  I moved the title assignment from the constructor, because if I got it there, it returned undefined to me. But if I get it in setTitle, it returns title

src/containers/Header/BreadcrumbsRouter/Breadcrumbs.js
```diff
@@ -33,6 +33,13 @@ const Breadcrumbs = ({
 
 	if (crumbs.length == 0) return null;
 
+	useEffect(() => {
+		// Adding subtitle to the title of the application
+		if (app.Services.TitleService != null) {
```
Pe5h4, line 38: @Aringoaway `if (app.Services.TitleService != undefined)`

src/services/TitleService.js
Pe5h4: @Aringoaway Yes, that is correct. It should not be in the constructor. However, it should be in `initialize` method, because you dont need to request the title everytime again and again. Its totally fine to get it once during the initialization.

```
initialize() {
this.title = this.App.Config.get("title");
}
```

Pe5h4 (changes requested):

src/services/TitleService.js
```diff
@@ -0,0 +1,16 @@
+import Service from '../abc/Service';
+export default class TitleService extends Service {
+	constructor(app, serviceName="TitleService"){
+		super(app, serviceName);
+		this.App = app;
```
Pe5h4, line 5: @Aringoaway Ok, I thought you will declare the `this.title` variable in constructor, which you should. I think it should be an empty string. `this.title = ""`

Pe5h4 (changes requested):

PR #426, opened 2023-06-06
Display button titles

aringocode:
### Changes
1. disabling the inputs
2. disabled the ability to press tree-menu-item and piechart parts after the end of the session
3. added page reloading, if the user presses the forward or back button in the browser
4. update CHANGELOG

https://github.com/TeskaLabs/asab-webui/assets/65866093/c0293a35-53be-4b37-ba02-3ac81461b835

### DIFF
--- a/src/modules/auth/index.js
+++ b/src/modules/auth/index.js
@@ -280,7 +280,7 @@ export default class AuthModule extends Module {
 			if (that.App.Store != null) {
 				that.App.Store.dispatch({ type: types.AUTH_SESSION_EXPIRATION, sessionExpired: true });
 				// Disable buttons and pagination in the whole screen
-				[...document.querySelectorAll('[class^="btn"]:not(.alert-button), [class*=" btn"]:not(.alert-button), .page-item, input')].forEach(i => {
+				[...document.querySelectorAll('[class^="btn"]:not(.alert-button), [class*=" btn"]:not(.alert-button), .btn-group a, .page-item, input')].forEach(i => {
 					i.classList.add("disabled");
 					i.setAttribute("disabled", "");
 				});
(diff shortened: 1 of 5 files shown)


src/modules/auth/index.js
```diff
@@ -280,14 +280,27 @@ export default class AuthModule extends Module {
...
 				// Disable link without class "nav-link" in the whole screen
 				[...document.querySelectorAll('a:not(.nav-link)')].forEach(i => {
 					i.classList.add("disabled-link");
 				});
+				// Disabling the ability to click on parts of widgets
+				[...document.querySelectorAll('.recharts-surface')].forEach(i => {
+					i.style.pointerEvents = "none"
+				});
+				// Disabling the ability to click on item in the Library
+				[...document.querySelectorAll('.tree-menu-item')].forEach(i => {
```
Pe5h4, line 296: @aringocode not sure about this, since what if we decide to change the library of tree menu?

src/modules/auth/index.js
```diff
@@ -280,14 +280,27 @@ export default class AuthModule extends Module {
...
+				[...document.querySelectorAll('[class^="btn"]:not(.alert-button), [class*=" btn"]:not(.alert-button), .btn-group a, .page-item, input, select')].forEach(i => {
 					i.classList.add("disabled");
 					i.setAttribute("disabled", "");
 				});
 				// Disable link without class "nav-link" in the whole screen
 				[...document.querySelectorAll('a:not(.nav-link)')].forEach(i => {
 					i.classList.add("disabled-link");
 				});
+				// Disabling the ability to click on parts of widgets
+				[...document.querySelectorAll('.recharts-surface')].forEach(i => {
```
Pe5h4, line 292: @aringocode Same here, what if we decide not to use recharts?

Pe5h4 (changes requested):
@aringocode It does not seem correct to me to have "baked" in some specific style references to external libraries

src/modules/auth/index.js
aringocode: I agree that these settings are specific to certain libraries. I deleted these parts, so now if the user wants to click on them, the page will reload)

aringocode: https://github.com/TeskaLabs/asab-webui/assets/65866093/1cea46ec-0bed-4ed1-99d3-72af25c38a98

Pe5h4 (approved):

PR #430, opened 2023-06-26
created preview flag component

queuing4oranges:
- created a reusable component which displays a preview flag (can be cutomized - here, name is set to "Preview")
Screenshot from 2023-06-26 21-53-35
- initial state is undefined
Screenshot from 2023-06-26 21-55-08


src/containers/Header/HelpButton.js
Pe5h4, line 1: @queuing4oranges Same here, dont include empty files

src/containers/Header/PreviewFlag.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+export default function PreviewFlag({ name }) {
+
+    if(!name) {
+        return null
+    }
+
+  return (
```
Pe5h4, line 10: @queuing4oranges Please update indentation. We use `tabs` which have a size of 4 spaces

src/containers/Header/header.scss
```diff
@@ -27,6 +27,12 @@ $toggle-text-color: var(--toggle-text-color);
 	//	overflow: hidden;
 	//	border: 0;
 	//}
+}
 
+.preview-flag {
+	font-size: 1rem;
+	color: $toggle-text-color;
+	padding-right: 1.5rem;
```
Pe5h4, line 35: @queuing4oranges Better to use margin here

src/containers/Header/header.scss
Pe5h4, line 1: @queuing4oranges I know its a legacy code, but can you please update the fontsize of the `help-button` to `1rem` ? Now its set to `1.1rem`

src/containers/Header/index.js
```diff
@@ -39,6 +41,7 @@ export function Header(props) {
 		<header className={`application-header ${headerProperties ? 'header-props-open' : ""}`}>
 			{(props.app.props.hasSidebar || typeof props.app.props.hasSidebar == 'undefined') ?
 				<>
+
```
Pe5h4, line 44: @queuing4oranges I think this empty line is not necessary here, remove it please

Pe5h4 (changes requested):
@queuing4oranges Good job, only minor refactoring request + please, add a documentation of how to use the setFlag component. Here is the example of how it can be done https://github.com/TeskaLabs/asab-webui/blob/master/doc/spinner.md

It should be in the `doc` section and written in markdown

src/containers/Header/PreviewFlag.js
Pe5h4: and apply it to the whole file ,)

doc/previewflag.md
```diff
@@ -0,0 +1,18 @@
+# ASAB WebUI PreviewFlag component
+
+`PreviewFlag` is a react reusable component.
+It is used to display a customizable (preview) flag. 
+
+# Setup
+
+To use the preview flag, pass the `name` prop to the component.
```
Pe5h4, line 8: @queuing4oranges It is also good to mention, that the `name` should be of `string` type

doc/previewflag.md
```diff
@@ -0,0 +1,18 @@
...
+
+# Usage
+
+```
+import { PreviewFlag } from 'asab-webui';
```
Pe5h4, line 13: @queuing4oranges I believe that the usage and import is a bit different. What we want to teach devs is to use the method/component for adding a custom flag correctly, so here it will be something like:

```
function SomeContainer(props){
	...

	props.app.setFlag(t("Preview"));
	
	...
}
```

This is I believe a correct usage. It will call the method `setFlag` from Application object and set the custom flag name for the particular container (this is intended behavior). If you will try it the way you have described, it will likely fail.

src/containers/Application.js
```diff
@@ -473,6 +473,21 @@ class Application extends Component {
 		}, [name])
 	}
 
+	setFlag(name) {
```
Pe5h4, line 476: @queuing4oranges I have an idea, that we may actually make the separate method component out of it, imported directly from asab-webui. We can try that together and validate, if the suggestion will be caught in infinite "too many renders" loop, or not :D

Pe5h4 (changes requested):
Lets have a small workshop on this to validate some ideas and we will see how it will go :)

Pe5h4 (changes requested):
@queuing4oranges Ok, it looks very good and I think it can be merged, but only after you will remove those 2 files you have commited https://github.com/TeskaLabs/asab-webui/pull/430/files#diff-ff6ca556837162fc8f73f5ac8224626700261870ca6879c8d565397413202e3b
 https://github.com/TeskaLabs/asab-webui/pull/430/files#diff-d26ff1bd605466c35705896ae9d483340c31ed1a8c6843b4d39b4ff114bd67e9
 
 - not sure why it appears here as an empty file, in my point of view, it should not be here at all :)

queuing4oranges:
- issues with committing empty files
- new PR instead of this one is #433

PR #431, opened 2023-06-29
Feature/bootstrap v5

petrKavulok:


src/modules/maintenance/ConfigModule/ConfigContainers/ConfigEditor.js
```diff
@@ -662,7 +662,7 @@ function ConfigSection(props) {
 					</div>
 				</Col>
 			</Row>
-			<FormGroup tag="fieldset" disabled={props.isSubmitting}>
```
Pe5h4, line 665: @petrKavulok Why? I thought that `<FormGroup>` is a reactstrap tag, or were there any clashes with bootstrap5?

petrKavulok: Hi, it `<FormGroup>` is using Bootstraps class names and respective styling. Form specific layout classes were dropped in B5. 
Screenshot 2023-08-09 at 9 57 48

https://getbootstrap.com/docs/5.1/migration/#forms

Pe5h4: Yes, but we are using reactstrap component here. And it is still present in reactstrap v9, which should be fully compatible with bootstrap v5

PR #435, opened 2023-07-07
Refactor/update service names

queuing4oranges:
- renaming seacat_auth, asab_config, lmio_remote_control, seacat_auth_webui to corresponding names with dashes instead
- list of used service in UI checked (no other used in asab-webui)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ module.exports = {
 		SERVICES: {
 			openidconnect: 'openidconnect',
 			asab-config: 'asab-config',
-			seacat_auth_webui: 'http://localhost:3000/auth'
+			seacat-auth-webui: 'http://localhost:3000/auth'
 			}
 		},
 	devConfig: {
@@ -159,7 +159,7 @@ module.exports = {
 		SERVICES: {
 			openidconnect: 'openidconnect',
 			asab-config: 'asab-config',
-			seacat_auth_webui: 'https://example.com/auth'
+			seacat-auth-webui: 'https://example.com/auth'
 			}
 		}
 }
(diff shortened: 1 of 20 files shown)


README.md
```diff
@@ -101,8 +101,8 @@ module.exports = {
 		API_PATH: 'api',
 		SERVICES: {
 			openidconnect: 'openidconnect',
-			asab_config: 'asab_config',
-			seacat_auth_webui: 'http://localhost:3000/auth'
+			asab-config: 'asab-config',
+			seacat-auth-webui: 'http://localhost:3000/auth'
```
Pe5h4, line 105: @queuing4oranges Since you've changed docs and readme, you should also change it here:

https://github.com/TeskaLabs/asab-webui/blob/refactor/update-service-names/src/modules/auth/header.js#L40-L41

:)

Pe5h4 (changes requested):

queuing4oranges:
yes. i will work on all applications, then merge it at once. there needs to be some changes in nginx, so i will have to tell everyone to adapt their configuration. 

for the service names i was using the list you provide in http://gitlab.teskalabs.int/team/plum/-/issues/1717
is it sufficient or could there still be other serivces?

Pe5h4:
> yes. i will work on all applications, then merge it at once. there needs to be some changes in nginx, so i will have to tell everyone to adapt their configuration.
> 
> for the service names i was using the list you provide in http://gitlab.teskalabs.int/team/plum/-/issues/1717 is it sufficient or could there still be other serivces?

Should be sufficient

queuing4oranges:
i found one more service from teska wiki. could you please have a quick look? :)

Pe5h4 (approved):
Yeah, it looks ok to me

PR #442, opened 2023-07-31
corrected locales

queuing4oranges:
- change the translation key to be reusable (there is no "AnalysisContainer" in bitswan, for example)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,8 @@
 
 ### Bugfix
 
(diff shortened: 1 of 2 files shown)


queuing4oranges:
coming to think about it, I'm actually not sure if should just update the name of the component (key) for the locales in each app. or make it more generic like above?

Pe5h4 (approved):
@queuing4oranges Yes, this is a good update

